PR #1, opened 2025-06-18
Manoj/sync invocation

manojdbos:
Most basic:

Sync invocation of workflow
updates to workflow_status

### DIFF
--- a/src/main/java/dev/dbos/transact/interceptor/TransactInvocationHandler.java
+++ b/src/main/java/dev/dbos/transact/interceptor/TransactInvocationHandler.java
@@ -46,8 +46,6 @@ public class TransactInvocationHandler implements InvocationHandler {
         logger.info("Interceptor called for method: " + method.getName());
 
         Method targetMethod = target.getClass().getMethod(method.getName(), method.getParameterTypes());
-        System.out.println("Has @Workflow: " +
-                targetMethod.isAnnotationPresent(Workflow.class));
 
         if (targetMethod.isAnnotationPresent(Workflow.class)) {
             return handleWorkflow(method, args, targetMethod.getAnnotation(Workflow.class));


README.md
```diff
@@ -1,6 +1,21 @@
 # dbos-transact-java
 DBOS Transact Java SDK 
 
+## Setting up dev environment
+
+Install a recent OpenJDK. I use OpenJDK 21.    
+https://adoptium.net/en-GB/temurin/releases/?os=any&arch=any&version=21
+
+Recommended IDE IntelliJ (Community edition is fine).
+But feel free to use vi, if you are more comfortable with it.
```
devhawk: Or VS Code! :)

src/main/java/dev/dbos/transact/migration/DatabaseMigrator.java
qianl15, line 1: nit: this file path name should be `migrations` instead of `migration` because we're going to have a lot more migrations in the future.

src/main/resources/db/migrations/V1__Create_DBOS_schema_tables.sql
```diff
@@ -1,29 +1,37 @@
...
+    inputs text,
+    started_at_epoch_ms bigint,
+    deduplication_id text,
+    priority integer NOT NULL DEFAULT 0,
+    UNIQUE (queue_name, deduplication_id)
 );
 
+CREATE INDEX workflow_status_created_at_index ON dbos.workflow_status USING btree (created_at);
+CREATE INDEX workflow_status_executor_id_index ON dbos.workflow_status USING btree (executor_id);
+CREATE INDEX workflow_status_status_index ON dbos.workflow_status USING btree (status);
```
kraftp, line 33: Let's remove the tables that aren't used from this file (workflow inputs and workflow events)

src/main/java/dev/dbos/transact/database/SystemDatabase.java
```diff
@@ -0,0 +1,606 @@
...
+        public String getStatus() {
+            return status;
+        }
+
+        public Long getDeadlineEpochMS() {
+            return deadlineEpochMS;
+        }
+    }
+
+    public static class UpdateWorkflowOptions {
```
kraftp, line 654: I don't understand what this big class is?

src/test/java/dev/dbos/transact/workflow/SimpleServiceImpl.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SimpleServiceImpl implements SimpleService {
+
+    Logger logger = LoggerFactory.getLogger(SimpleServiceImpl.class);
+
+    @Workflow(name = "workWithString")
+    public String workWithString(String input) {
```
kraftp, line 13: To make sure I'm clear: later on, we'll make it so workflows always return handles?

src/main/java/dev/dbos/transact/database/SystemDatabase.java
manojdbos: It is the same UpdateWorkFlowOPtions from TS and python for the updateWorkflowStatus method. The update sql is generated based on options provided.

Large ness is due to Java verbosity with get/set methods

src/test/java/dev/dbos/transact/workflow/SimpleServiceImpl.java
manojdbos: I have thought about how it will be done. We can talk about it

src/main/resources/db/migrations/V1__Create_DBOS_schema_tables.sql
manojdbos: workflow_events will be needed when we do events right ? Or No ?

manojdbos: removed workflow_inputs

PR #2, opened 2025-06-20
Manoj/async

manojdbos:


src/main/java/dev/dbos/transact/database/SystemDatabase.java
```diff
@@ -428,6 +425,22 @@ public void updateWorkflowStatus(
...
+                return output.get(0);
+            }
+        } catch (SQLException e) {
+            logger.error("Error retrieving workflow for "+workflowId, e);
+        }
+
+        throw new NonExistentWorkflowException(workflowId) ;
+}
+
 public List<WorkflowStatus> getWorkflows(GetWorkflowsInput input) throws SQLException {
```
qianl15: Nit: would be better to rename this `listWorkflows` and `ListWorkflowsInput`

src/main/java/dev/dbos/transact/interceptor/AsyncInvocationHandler.java
```diff
@@ -0,0 +1,86 @@
...
+
+
+    public AsyncInvocationHandler(Object target, DBOSExecutor dbosExecutor) {
+        this.target = target;
+        this.targetClassName = target.getClass().getName();
+        this.dbosExecutor = dbosExecutor ;
+    }
+
+    @Override
+    public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
```
qianl15, line 46: Why is the `proxy` not used in this function?

src/test/java/dev/dbos/transact/workflow/AsyncWorkflowTest.java
```diff
@@ -0,0 +1,176 @@
...
+    @Test
+    public void sameWorkflowId() throws SQLException  {
+
+        SimpleAsyncService simpleAsyncService = dbos.<SimpleAsyncService>Workflow()
+                .interfaceClass(SimpleAsyncService.class)
+                .implementation(new SimpleServiceImpl())
+                .async()
+                .build();
+
+        SimpleServiceImpl.executionCount =0 ;
```
qianl15: Nit: is there a Java code formatter that we can use to clean up our code format? In Python we use Black, TS uses prettier.

src/test/java/dev/dbos/transact/workflow/SimpleServiceImpl.java
```diff
@@ -21,4 +21,6 @@ public void workWithError() throws Exception {
        throw new Exception("DBOS Test error") ;
     }
 
+
```
qianl15, line 27: In this file, looks like `SimpleServiceImpl` implements `SimpleService` but not `SimpleAsyncService`. Will Java report a compiler error if we write `SimpleServiceImpl implements SimpleAsyncService` because the interface doesn't match with the implementation (due to WorkflowHandle)?

src/main/java/dev/dbos/transact/interceptor/AsyncInvocationHandler.java
qianl15, line 1: As we discussed, for now we can merge the sync and async paths, making every workflow invocation in an executor pool. This will hopefully simplify the interface a lot.

src/main/java/dev/dbos/transact/interceptor/AsyncInvocationHandler.java
manojdbos: That object is the proxy object on which the method was called.
In most cases it can be ignored. It can be useful in some special cases like needing to pass the proxy as a callback or if the the invoke method needs to call another method on the proxy.

So far I have not needed it

src/test/java/dev/dbos/transact/workflow/SimpleServiceImpl.java
manojdbos: I will change the test case impl class to not implement any interface so that it is clear

src/test/java/dev/dbos/transact/workflow/AsyncWorkflowTest.java
manojdbos: will look for one

src/test/java/dev/dbos/transact/workflow/SimpleServiceImpl.java
manojdbos: made the naming clearer

src/main/java/dev/dbos/transact/interceptor/AsyncInvocationHandler.java
manojdbos: fixed

PR #3, opened 2025-06-24
Async take 2

manojdbos:
For async invocation , client required to ignore the return value

### DIFF
--- a/src/main/java/dev/dbos/transact/interceptor/AsyncInvocationHandler.java
+++ b/src/main/java/dev/dbos/transact/interceptor/AsyncInvocationHandler.java
@@ -44,8 +44,7 @@ public class AsyncInvocationHandler implements InvocationHandler {
 
     @Override
     public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
-
-        // Method targetMethod = findMethodWithSameSignature(target.getClass(), method);
+        
         Method implMethod = target.getClass().getMethod(method.getName(), method.getParameterTypes());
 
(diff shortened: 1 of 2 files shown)


src/main/java/dev/dbos/transact/interceptor/AsyncInvocationHandler.java
```diff
@@ -0,0 +1,89 @@
...
+
+            dbosExecutor.submitWorkflow(
+                    workflowName,
+                    targetClassName,
+                    method.getName(),
+                    args,
+                    () -> (Object) method.invoke(target, args)
+            );
+
+            return null ; // always return null or default
```
qianl15: What happens if the return type is a primitive type? Will this throw an error?

src/main/java/dev/dbos/transact/interceptor/TransactInvocationHandler.java
```diff
@@ -12,6 +12,11 @@
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
 
+/**
+ *
+ * To be removed: See AsyncInvocationHandler
```
qianl15: We might want to keep this for now. Separating sync from async.

src/main/java/dev/dbos/transact/interceptor/AsyncInvocationHandler.java
manojdbos: was going to add the code. need to return a default value

PR #4, opened 2025-06-26
Manoj/steps Minimal steps implementation

manojdbos:
read and write operation_outputs
list steps
Workflow can invoke steps

### DIFF
--- a/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
+++ b/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
@@ -202,8 +202,7 @@ public class DBOSExecutor {
             return result ;
         };
 
-        /*ContextAwareCallable<T> contextAwareTask = new ContextAwareCallable<>(task);
-        contextAwareTask.setWorkflowId(DBOSContextHolder.get().getWorkflowId()); */
+        // Copy the context - dont just pass a reference - memory visibility
         ContextAwareCallable<T> contextAwareTask = new ContextAwareCallable<>(DBOSContextHolder.get().copy(),task);
         Future<T> future = executorService.submit(contextAwareTask);
 


src/main/java/dev/dbos/transact/database/SystemDatabase.java
```diff
@@ -438,214 +442,268 @@ public WorkflowStatus getWorkflow(String workflowId) {
...
+
+
+        return workflows ;
     }
 
+    /**
+    *  Helper method for tests
+     *  Should be moved to TestUtils
+     */
+    public void deleteWorkflowsTestHelper() throws SQLException{
```
kraftp, line 604: Even better, replace this with a generic "reset system database" method that drops the entire database and can be used in our tests and exposed to help users write their own tests. We have something similar in TS + Python.

src/main/java/dev/dbos/transact/database/SystemDatabase.java
```diff
@@ -438,214 +442,268 @@ public WorkflowStatus getWorkflow(String workflowId) {
...
 
-/**
-*  Helper method for tests
- *  Should be moved to TestUtils
- */
-public void deleteWorkflowsTestHelper() throws SQLException{
+    public void deleteOperations() throws SQLException{
 
-    String sql = "delete from dbos.workflow_status";
+        String sql = "delete from dbos.operation_outputs;";
```
kraftp, line 623: Then we wouldn't need this if we're dropping the whole database.

src/test/java/dev/dbos/transact/step/ServiceWFAndStepImpl.java
```diff
@@ -0,0 +1,31 @@
+package dev.dbos.transact.step;
+
+import dev.dbos.transact.workflow.Step;
+import dev.dbos.transact.workflow.Workflow;
+
+public class ServiceWFAndStepImpl implements ServiceWFAndStep {
+
+    private ServiceWFAndStep self ;
+
+    public void setSelf(ServiceWFAndStep serviceWFAndStep) {
```
kraftp, line 10: What is this?

src/test/java/dev/dbos/transact/step/ServiceA.java
```diff
@@ -0,0 +1,10 @@
+package dev.dbos.transact.step;
```
kraftp, line 1: I understand it's not idiomatic Java, but is it possible for the test classes to be co-located with their tests as in the other languages? It will be easier to write tests in the future if they aren't spread across many files.

src/main/java/dev/dbos/transact/workflow/internal/StepResult.java
```diff
@@ -0,0 +1,31 @@
+package dev.dbos.transact.workflow.internal;
+
+public class StepResult {
```
kraftp, line 3: I don't understand what is the difference between this and StepInfo?

kraftp: I see, this is a purely internal thing?

src/test/java/dev/dbos/transact/step/ServiceWFAndStepImpl.java
manojdbos: When you call a step from a workflow , the call needs to go through a proxy. The method can be named anything. It is setSelf because I have called the variable self.

In the other test, there were 2 services. There was setServiceB

qianl15: My understanding is this is a way to set `self` to a proxy. This way, workflows can invoke its own steps.

src/main/java/dev/dbos/transact/database/SystemDatabase.java
manojdbos: I will develop this as we go along

src/main/java/dev/dbos/transact/workflow/internal/StepResult.java
manojdbos: There is none. But in both TS and Python , there is internal calls like OperationResultInternal and a public StepInfo ... which are both the same

manojdbos: But you are right. We could just keep one

src/test/java/dev/dbos/transact/step/ServiceA.java
manojdbos: The maven and gradle build expect things in certain directories. The directory structure is dictated by that and every  java developer is used to it. it can be changed but it would be hard for us and for any user who tries to get the code and build it

PR #5, opened 2025-06-27
Manoj/recovery 2

manojdbos:
WorkflowRegistry key = name , value = WorkflowFunctionWrapper
invoke uses workflow registry
executeById
recovery method

### DIFF
--- a/src/test/java/dev/dbos/transact/execution/DBOSExecutorTest.java
+++ b/src/test/java/dev/dbos/transact/execution/DBOSExecutorTest.java
@@ -199,13 +199,13 @@ class DBOSExecutorTest {
             result = executingService.workflowMethodWithStep("test-item");
         }
 
-        assertEquals("test-itemstepOne", result);
+        assertEquals("test-itemstepOnestepTwo", result);
 
         List<WorkflowStatus> wfs = systemDatabase.listWorkflows(new ListWorkflowsInput()) ;
         assertEquals(wfs.get(0).getStatus(), WorkflowState.SUCCESS.name());
 
         List<StepInfo> steps = systemDatabase.listWorkflowSteps(wfid) ;
-        assertEquals(1, steps.size());
+        assertEquals(2, steps.size());
 
         setWorkflowState(dataSource, wfid, WorkflowState.PENDING.name());
         deleteSteps(dataSource, wfid);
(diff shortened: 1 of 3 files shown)


src/test/java/dev/dbos/transact/execution/DBOSExecutorTest.java
```diff
@@ -0,0 +1,275 @@
...
+
+       ExecutingService executingService = dbos.<ExecutingService>Workflow()
+                .interfaceClass(ExecutingService.class)
+                .implementation(new ExecutingServiceImpl())
+                .build();
+
+        String result = null ;
+
+        String wfid = "wf-123";
+        try (SetWorkflowID id = new SetWorkflowID(wfid)){
```
qianl15, line 92: How is the `id` variable used? I think this semantics looks a bit non-intuitive because it's creating a new variable but not used anywhere. If it's used somewhere, can we have a test for it?

src/test/java/dev/dbos/transact/execution/DBOSExecutorTest.java
```diff
@@ -0,0 +1,275 @@
...
+
+        wfs = systemDatabase.listWorkflows(new ListWorkflowsInput()) ;
+        assertEquals(wfs.get(0).getStatus(), WorkflowState.SUCCESS.name());
+        steps = systemDatabase.listWorkflowSteps(wfid) ;
+        assertEquals(1, steps.size());
+
+    }
+
+
+    private void setWorkflowState(DataSource ds, String workflowId, String newState) throws SQLException {
```
kraftp, line 280: Can some of these helper methods be factored out and reused across test files?

src/test/java/dev/dbos/transact/utlils/DBUtils.java
```diff
@@ -0,0 +1,87 @@
+package dev.dbos.transact.utlils;
```
kraftp: Spelling

src/test/java/dev/dbos/transact/execution/DBOSExecutorTest.java
```diff
@@ -0,0 +1,275 @@
...
+        }
+
+        assertTrue(error);
+
+    }
+
+    @Test
+    public void executeWithStep() throws Exception {
+
+        ExecutingService executingService = dbos.<ExecutingService>Workflow()
```
kraftp, line 187: Can we update the test to recover _from_ a step? Stop the workflow after step one then recover, verifying that it completes step two without re-executing step one.

src/test/java/dev/dbos/transact/execution/DBOSExecutorTest.java
manojdbos: It is not used. Will check if there is a better syntax ?

src/test/java/dev/dbos/transact/execution/DBOSExecutorTest.java
manojdbos: Yes. I move methods to the utils folder as soon as they need to be reused

src/test/java/dev/dbos/transact/execution/DBOSExecutorTest.java
manojdbos: Added a test

src/test/java/dev/dbos/transact/execution/DBOSExecutorTest.java
manojdbos: Some other options are 

(1) Another better name like
try (WorkflowContextScope scope = WorkflowContextScope.withWorkflowId(wfid)) {
    result = executingService.workflowMethodWithStep("test-item");
}
However scope variable is still ignored

(2) Another option is to use a Runnable or Callable 

DBOSContext.runWithWorkflowId(wfid, () -> {
    result = executingService.workflowMethodWithStep("test-item");
});


This has the feel of async and gets messier when you need run child workflows

(3) Explicit -- use has to call methods to do it . Too error prone for the user

String prev = DBOSContext.get().getWorkflowId();
DBOSContext.get().setWorkflowId(wfid);
try {
    result = executingService.workflowMethodWithStep("test-item");
} finally {
    DBOSContext.get().setWorkflowId(prev); // Restore
}

I think what we have with maybe better naming -- is most convenient when you have nested scopes

PR #6, opened 2025-07-02
Queues

manojdbos:
Basic queues implementation

Not Included:
Enque option -Deduplication
Enque option - priority

### DIFF
--- a/src/main/java/dev/dbos/transact/database/QueuesDAO.java
+++ b/src/main/java/dev/dbos/transact/database/QueuesDAO.java
@@ -74,7 +74,7 @@ public class QueuesDAO {
             }
 
             // Calculate max tasks based on concurrency limits
-            int maxTasks = Integer.MAX_VALUE;
+            int maxTasks = 100;
 
             if (queue.getWorkerConcurrency() > 0 || queue.getConcurrency() > 0) {
                 // Count pending workflows by executor
@@ -147,8 +147,12 @@ public class QueuesDAO {
                 queryBuilder.append(" LIMIT ?");
             }
 
-            // Add FOR UPDATE NOWAIT
-            queryBuilder.append(" FOR UPDATE NOWAIT");
+            // Add FOR UPDATE NOWAIT or SKIP Locked
+            if (queue.getConcurrency() > 0 ) {
+                queryBuilder.append(" FOR UPDATE NOWAIT");
+            } else {
+                queryBuilder.append(" FOR UPDATE SKIP LOCKED");
+            }
 
             String workflowsQuery = String.format(queryBuilder.toString(), Constants.DB_SCHEMA);
 
--- a/src/main/java/dev/dbos/transact/queue/QueueService.java
+++ b/src/main/java/dev/dbos/transact/queue/QueueService.java
@@ -11,6 +11,7 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ThreadLocalRandom;
 
 import static java.lang.Math.max;
+import static java.lang.Math.min;
 
 public class QueueService {
 
@@ -78,6 +79,7 @@ public class QueueService {
 
                     } catch (Exception e) {
 
+                        pollingInterval = min(maxPollingInterval, pollingInterval*2);
                         logger.error("Error executing queued workflow", e);
                     }
 


src/main/java/dev/dbos/transact/database/QueuesDAO.java
```diff
@@ -0,0 +1,223 @@
...
+                queryBuilder.append(" ORDER BY created_at ASC");
+            }
+
+            // Add limit if not infinite
+            if (maxTasks != Integer.MAX_VALUE) {
+                queryBuilder.append(" LIMIT ?");
+            }
+
+            // Add FOR UPDATE NOWAIT
+            queryBuilder.append(" FOR UPDATE NOWAIT");
```
kraftp: Can you update this to reflect the latest Python? We've moved from NOWAIT to SKIP LOCKED to improve performance. A few other changes as well (https://github.com/dbos-inc/dbos-transact-py/pull/398).

src/test/java/dev/dbos/transact/queue/QueuesTest.java
```diff
@@ -0,0 +1,459 @@
...
+        dataSource = DBUtils.createDataSource(dbosConfig);
+        DBOS.initialize(dbosConfig);
+        dbos = DBOS.getInstance();
+        SystemDatabase.initialize(dataSource);
+        systemDatabase = SystemDatabase.getInstance();
+        dbosExecutor = new DBOSExecutor(dbosConfig, systemDatabase);
+        dbos.setDbosExecutor(dbosExecutor);
+        dbos.launch();
+        // TODO: move before launch after launch has a wait
+        queueService = new QueueService(systemDatabase);
```
kraftp, line 84: Shouldn't all this configuration be done `dbos.launch()`?

src/test/java/dev/dbos/transact/queue/QueuesTest.java
```diff
@@ -0,0 +1,459 @@
...
+
+        Queue firstQ = new DBOS.QueueBuilder("firstQueue")
+                .concurrency(1)
+                .workerConcurrency(1)
+                .build();
+
+        ServiceQ serviceQ = new DBOS.WorkflowBuilder<ServiceQ>()
+                .interfaceClass(ServiceQ.class)
+                .implementation(new ServiceQImpl())
+                .queue(firstQ)
```
kraftp, line 144: It looks like to use queues we'll have to create another proxy of the class configured with the queue?

It's fine for now to get started, but we need to think of a better interface, or you'll need a dozen proxies to do anything (one for sync, one for queue A, one for queue B...).

src/test/java/dev/dbos/transact/queue/QueuesTest.java
qianl15: I see there is already `queueService = new QueueService` in dbos launch()

src/test/java/dev/dbos/transact/queue/QueuesTest.java
```diff
@@ -0,0 +1,459 @@
...
+        dbos.setQueueService(queueService);
+        queueService.start();
+
+
+        DBUtils.clearTables(dataSource);
+    }
+
+    @AfterEach
+    void afterEachTest() throws SQLException {
+        queueService.stop();
```
qianl15, line 95: Same as above, queue service start and stop should be part of `dbos.launch()` and `dbos.shutdown()`

src/test/java/dev/dbos/transact/queue/QueuesTest.java
manojdbos: Yes launch does it.
For tests configuring each components and assembling DBOS lets you write better and simpler tests. End users will never do it. Eventually not all tests need this assembling. Some tests can just use launch
Also launch needs little more work.

src/test/java/dev/dbos/transact/queue/QueuesTest.java
manojdbos: Yes . There are a few other options that can be investigated

(1) We can provide a lambda based startWorkflow style method
(2) Keep one type of proxy but provide queueparams as context params

src/test/java/dev/dbos/transact/queue/QueuesTest.java
manojdbos: For testing, it better to have finer control. 

In the end we can have a few tests that do things just with dbos.launch().

src/test/java/dev/dbos/transact/queue/QueuesTest.java
manojdbos: If you look at launch, it will start the sub components when they are null.

src/test/java/dev/dbos/transact/queue/QueuesTest.java
manojdbos: In fact for 2 of the tests, I need to temporarily shutdown the queueservice thread but not the entire dbos instance.

PR #7, opened 2025-07-05
ChildWorkflows

manojdbos:
ChildWorkflows
Context with rest of fields
Refactored Inovcation Handlers

### DIFF
--- a/src/main/java/dev/dbos/transact/interceptor/BaseInvocationHandler.java
+++ b/src/main/java/dev/dbos/transact/interceptor/BaseInvocationHandler.java
@@ -69,14 +69,12 @@ public abstract class BaseInvocationHandler implements InvocationHandler {
             // child workflow
             if (ctx.hasParent()) {
                 // child called with SetWorkflowId
-                logger.info("child with set") ;
                 result = submitWorkflow(workflowName, targetClassName, wrapper, args);
             } else {
                 // child called without Set
                 // create child context from the parent
 
-                String childId = ctx.getWorkflowId() + "_" + ctx.getParentFunctionId();
-                logger.info("child call ......" + childId);
+                String childId = ctx.getWorkflowId() + "-" + ctx.getParentFunctionId();
                 try(SetWorkflowID id = new SetWorkflowID(childId)) {
                     result = submitWorkflow(workflowName, targetClassName, wrapper, args);
                 }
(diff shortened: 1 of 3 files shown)


src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -180,29 +184,46 @@ public void launch() {
...
 
-    }
-
-    public void shutdown() {
-        queueService.stop();
-        dbosExecutor.shutdown();
-        instance = null ;
-    }
-
+        // Block the main thread until shutdown is called
```
qianl15, line 187: I think we might want to remove this blocker thread. We don't have it in Python/TS -- because I think the library shouldn't block the code from exiting. If a user sets up an HTTP server we can hook `DBOS.shutdown()` to the server shutdown. Otherwise, users should have full control over the behavior. For example, they can choose how long to run the server and how to handle interrupted exceptions.

src/main/java/dev/dbos/transact/exceptions/DBOSAppException.java
```diff
@@ -0,0 +1,17 @@
+package dev.dbos.transact.exceptions;
+
+
+/**
+ * Wrapper to be thrown back to user from calls like getResult
+ * It would be unsafe, error-prone and hard to try to throw the
+ * actual user defined exception
+ *
+ */
+public class DBOSAppException extends RuntimeException {
```
qianl15, line 10: Can we define which exceptions are serializable? Then only wrap the exception if we cannot serialize it?

qianl15: My worry is that users would see different exceptions when working with DBOS. Might not be a huge deal as long as we define the behavior well.

src/main/java/dev/dbos/transact/interceptor/BaseInvocationHandler.java
```diff
@@ -0,0 +1,148 @@
...
+            // child workflow
+            if (ctx.hasParent()) {
+                // child called with SetWorkflowId
+                logger.info("child with set") ;
+                result = submitWorkflow(workflowName, targetClassName, wrapper, args);
+            } else {
+                // child called without Set
+                // create child context from the parent
+
+                String childId = ctx.getWorkflowId() + "_" + ctx.getParentFunctionId();
```
qianl15: Nit: we use `-` not underscore to append the function ID.
```suggestion
                String childId = ctx.getWorkflowId() + "-" + ctx.getParentFunctionId();
```

src/main/java/dev/dbos/transact/DBOS.java
manojdbos: Without the blocker thread , we might need provide a DBOS.wait -- Other wise main() might exit before their async code executes

src/main/java/dev/dbos/transact/exceptions/DBOSAppException.java
manojdbos: This approach is safer. But we can iterate over it. We'll need to impose some restriction on what exception can be throw.

PR #8, opened 2025-07-08
Scheduled Workflows

manojdbos:


src/main/java/dev/dbos/transact/json/Boxed.java
```diff
@@ -0,0 +1,14 @@
+package dev.dbos.transact.json;
+
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+public class Boxed {
```
qianl15, line 5: What's the purpose of this class? Based on the code it seems to work for serialization/deserialization. Does it work for primitives?

src/main/java/dev/dbos/transact/scheduled/SchedulerService.java
```diff
@@ -0,0 +1,115 @@
...
+        WorkflowFunctionWrapper wrapper = dbosExecutor.getWorkflow(workflowName);
+        if (wrapper == null) {
+            throw new IllegalStateException("Workflow not registered: " + workflowName);
+        }
+
+        Runnable scheduleTask = new Runnable() {
+            @Override
+            public void run() {
+                try {
+                    ZonedDateTime scheduledTime = ZonedDateTime.now();
```
qianl15: I think by default cron should use the UTC timezone. Does `ZonedDateTime.now()` return the local time or UTC time?

src/test/java/dev/dbos/transact/scheduled/EveryMinute.java
```diff
@@ -0,0 +1,15 @@
+package dev.dbos.transact.scheduled;
```
kraftp, line 1: It seems like not all the test classes are actually used in the tests? Which is fine, we only need a couple tests for this.

src/main/java/dev/dbos/transact/scheduled/SchedulerService.java
```diff
@@ -0,0 +1,115 @@
...
+                    Object[] args = new Object[2];
+                    args[0] = scheduledTime.toInstant();
+                    args[1] = ZonedDateTime.now().toInstant();
+                    logger.info("submitting to dbos Executor " + workflowName);
+                    String workflowId = String.format("sched-%s-%s",workflowName, scheduledTime.toString()) ;
+                    try(SetWorkflowID id = new SetWorkflowID(workflowId)) {
+                        dbosExecutor.submitWorkflow(workflowName, instance.getClass().getName(), wrapper.target, args, wrapper.function);
```
kraftp: We should make cron use queues (specifically, the internal queue) as it does in Python&mdash;this produces much simpler behavior in a distributed setting when many processes are trying to launch the same cron.

src/main/java/dev/dbos/transact/scheduled/SchedulerService.java
```diff
@@ -0,0 +1,115 @@
...
+import java.time.Instant;
+import java.time.ZonedDateTime;
+import java.util.Arrays;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+public class SchedulerService {
+
+    private final ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(4);
```
qianl15, line 30: I think the threadpool size needs to be larger. Otherwise it may cause deadlock when there are many scheduled tasks.

src/main/java/dev/dbos/transact/json/Boxed.java
kraftp, line 5: What's this do?

src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -225,5 +237,15 @@ public static WorkflowHandle retrieveWorkflow(String workflowId) {
         return DBOS.getInstance().dbosExecutor.retrieveWorkflow(workflowId);
     }
 
+    /**
+     * Scans the class for all methods that have Workflow and Scheduled annotations
+     * and schedules them for execution
+     *
+     * @param implementation instance of a class
+     */
+    public void scheduleWorkflow(Object implementation) {
```
kraftp, line 250: Can we call this in DBOS.launch instead of users needing to call it themselves?

src/main/java/dev/dbos/transact/json/Boxed.java
manojdbos: It is need for the serialization of certain types like arrays of primitives, certain collections , types like Instant

src/test/java/dev/dbos/transact/scheduled/EveryMinute.java
manojdbos: They are all used.

src/main/java/dev/dbos/transact/scheduled/SchedulerService.java
manojdbos: What size do you recommend ?

src/main/java/dev/dbos/transact/json/Boxed.java
manojdbos: See response above. Need for primitive arrays, certain collections, types like Instant

src/main/java/dev/dbos/transact/DBOS.java
manojdbos: That can happen in the future.

To do that, I would need to scan the code for all classes that have Workflow and scheduled annotations. I think Java supports component scanning. But we can add it later.

src/main/java/dev/dbos/transact/json/Boxed.java
manojdbos: type info is needed for serialization/deser.
This is the first attempt. Letting jackson do it mostly.
If there are issues, might need to store type info more explicitly

src/main/java/dev/dbos/transact/scheduled/SchedulerService.java
manojdbos: Fixed

src/main/java/dev/dbos/transact/scheduled/SchedulerService.java
kraftp: If we're using queues, there shouldn't be a thread pool here, everything goes on the queue

src/main/java/dev/dbos/transact/scheduled/SchedulerService.java
```diff
@@ -0,0 +1,125 @@
...
+        });
+    }
+
+    public void stop () {
+        stop = true ;
+    }
+
+    public void start () {
+        schedulerQueue = new DBOS.QueueBuilder("schedulerQueue")
+                .workerConcurrency(4) // are there reasonable defaults ?
```
kraftp: The internal queue should have no default concurrency limits

src/main/java/dev/dbos/transact/scheduled/SchedulerService.java
```diff
@@ -0,0 +1,125 @@
...
+                    e.printStackTrace();
+                }
+
+                if (!stop) {
+                    logger.info("Scheduling the next execution") ;
+                    ZonedDateTime now = ZonedDateTime.now(ZoneOffset.UTC);
+                    executionTime.nextExecution(now).ifPresent(nextTime -> {
+                        logger.info("Next execution time " + nextTime.toString());
+                        long delayMs = Duration.between(now, nextTime).toMillis();
+                        scheduler.schedule(this, delayMs, TimeUnit.MILLISECONDS);
```
kraftp, line 99: What does this do? We shouldn't need this thread pool any more now that we're using queues

src/main/java/dev/dbos/transact/scheduled/SchedulerService.java
manojdbos: That is the schedulingservice pool -- at the correct time it will run the code that enqueues the workflow. The use of a timer is not recommended

src/main/java/dev/dbos/transact/scheduled/SchedulerService.java
manojdbos: The pool is needed to schedule the task that enqueues the workflow. Java discourages the use of timers

src/main/java/dev/dbos/transact/scheduled/SchedulerService.java
manojdbos: removed

src/main/java/dev/dbos/transact/scheduled/SchedulerService.java
manojdbos: fixed

src/main/java/dev/dbos/transact/scheduled/SchedulerService.java
manojdbos: task is very short - enqueuing the workflow. So I think we do not need a large pool.

PR #11, opened 2025-07-15
Send And Recv API

manojdbos:


src/main/java/dev/dbos/transact/database/NotificationsDAO.java
```diff
@@ -0,0 +1,282 @@
...
+                       String topic, double timeoutSeconds) throws SQLException, InterruptedException {
+
+        String functionName = "DBOS.recv";
+        String finalTopic = (topic != null) ? topic : Constants.DBOS_NULL_TOPIC ;
+
+        // First, check for previous executions
+        StepResult recordedOutput = null ;
+
+        try (Connection c = dataSource.getConnection()) {
+            logger.info("mjjjj checking recorded output for recv" + workflowUuid + " " + functionId) ;
```
kraftp: Let's clean these up

src/main/java/dev/dbos/transact/notifications/NotificationService.java
```diff
@@ -0,0 +1,189 @@
...
+
+        notificationsMap.clear();
+        logger.info("Notification listener stopped");
+    }
+
+    public void send(String destinationId, Object message, String topic) {
+
+        DBOSContext ctx = DBOSContextHolder.get() ;
+        if (!ctx.isInWorkflow()) {
+            // TODO : temp workflow
```
kraftp, line 79: Hoping we don't have to do temp workflows in Go/Java. We'll see.

PR #13, opened 2025-07-17
Manoj/unifiedproxy

manojdbos:
DBOSOption

 DBOSOptions options = new DBOSOptions.Builder(wfid1).build();
        String result;
        try (SetDBOSOptions id = new SetDBOSOptions(options)){
            result = simpleService.workWithString("test-item");
        }

### DIFF
--- a/src/main/java/dev/dbos/transact/context/SetWorkflowID.java
+++ b/src/main/java/dev/dbos/transact/context/SetWorkflowID.java
@@ -1,9 +1,6 @@
 package dev.dbos.transact.context;
 
-/**
- * If we like SetDBOSOptions we should remove it and keep just API
- *
- */
+
 public class SetWorkflowID implements AutoCloseable {
     private final DBOSContext previousCtx ;
 


src/test/java/dev/dbos/transact/workflow/UnifiedProxyTest.java
```diff
@@ -0,0 +1,180 @@
...
+    @BeforeEach
+    void beforeEachTest() throws SQLException {
+        UnifiedProxyTest.dataSource = DBUtils.createDataSource(dbosConfig) ;
+        DBOS.initialize(dbosConfig);
+        dbos = DBOS.getInstance();
+        SystemDatabase.initialize(dataSource);
+        systemDatabase = SystemDatabase.getInstance();
+        dbosExecutor = new DBOSExecutor(dbosConfig, systemDatabase);
+        dbos.setDbosExecutor(dbosExecutor);
+        dbos.launch();
```
qianl15, line 73: I assume eventually we'll just need `DBOS.launch()` -- so users don't need to do all the set up themselves?

manojdbos: This is only for the test.

If you look at the sample apps, we only call dbos.launch.

PR #14, opened 2025-07-18
Manoj/tomcat

manojdbos:
(1) Support for http
(2) Embedded Tomcat
(3) AdminController

### DIFF
--- a/src/main/java/dev/dbos/transact/config/DBOSConfig.java
+++ b/src/main/java/dev/dbos/transact/config/DBOSConfig.java
@@ -107,18 +107,17 @@ public class DBOSConfig {
             return this;
         }
 
-        public Builder http() {
+        public Builder runAdminServer() {
             this.http = true ;
             return this;
         }
 
-        public Builder httpAdminPort(int port) {
+        public Builder adminServerPort(int port) {
             this.httpPort = port;
             return this;
         }
 
-
-        public Builder httpAwaitOnStart(boolean wait) {
+        public Builder adminAwaitOnStart(boolean wait) {
             this.httpAwaitOnStart = wait;
             return this;
         }
--- a/src/main/java/dev/dbos/transact/Main.java
+++ b/src/main/java/dev/dbos/transact/Main.java
@@ -14,8 +14,8 @@ public class Main {
                     .dbUser("postgres")
                     .sysDbName("dbos_java_sys")
                     .maximumPoolSize(2)
-                    .http()
-                    .httpAdminPort(8080)
+                    .runAdminServer()
+                    .adminServerPort(8080)
                     .build();
 
             DBOS.initialize(dbosConfig);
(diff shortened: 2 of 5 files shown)


src/main/java/dev/dbos/transact/config/DBOSConfig.java
```diff
@@ -94,6 +113,31 @@ public Builder sysDbName(String sysDbName) {
             return this;
         }
 
+        public Builder http() {
```
kraftp: I don't understand what all this configuration is for. If it's just the admin server, we should have the same two options as in TS/Python (runAdminServer and adminServerPort) and provide no further configuration as the admin server is a purely internal thing. If it's more than just the admin server, I don't understand its purpose. We shouldn't be trying to do HTTP for users, plenty of ways to do that already.

manojdbos: This is for just turning on the httpServer or off it is not called.
Without it no DBOS httpServer. I think we need this when people bring their own httpServer.
Our AdminController is Jersey compliant. When people bring their httpServer , we recommend
jersey complaint servers ( I think almost all know java httpservers are Jersey complaint) . Our
admin controller should work with their server with a little registration.
SpringBoot also does Spring Controllers which have slightly different annotations. We can provide a wrapper.

But most importantly, I think when people bring their http Server, we can just shut down ours by not caling http(). There is programmatic access to the admin API ( or we can provide it if it is not there). This is another way to approach it.

Admin port is right now hardcoded. I can provide a config field for it.
httpPort is the port for user controllers.

kraftp: I don't think DBOS needs to know about a user HTTP server.

The admin server is separate and purely internal and should not be affected by any user HTTP server.

manojdbos: If we say, everyone must bring their httpServer. Why do we even need a httpServer ? the only reason is the cloud health check.

kraftp: We need an admin API for the cloud to use. It hosts not just the health check but also important endpoints like the recovery and workflow management/observability endpoints. It should be totally internal and separate from the user's app and server.

manojdbos: Fixed.
Now default is http off
http() -- turns on admin http for port 3001
httpAdminPort() -- let you set the admin port

removed packages and controllerInstances from config since they will now never be required

manojdbos: I will update the toolbox REST demo to create a 2nd tomcat in the user space -- as an example .

kraftp: Awesome! But let's call them `runAdminServer` and `adminServerPort` so it's unambiguous what these do.

PR #15, opened 2025-07-22
Manoj/sleep

manojdbos:


src/test/java/dev/dbos/transact/execution/DBOSExecutorTest.java
```diff
@@ -275,6 +276,78 @@ public void ReExecuteWithStepTwoOnly() throws Exception {
...
+        long start = System.currentTimeMillis() ;
+        try (SetWorkflowID id = new SetWorkflowID(wfid)){
+            executingService.sleepingWorkflow(.002f);
+        }
+
+        List<StepInfo> steps = systemDatabase.listWorkflowSteps(wfid);
+
+        assertEquals("DBOS.sleep", steps.get(0).getFunctionName());
+
+        // let us set the state to PENDING and increase the sleep time
```
kraftp, line 332: This seems too complex for a test--why do we need to change the internal values? What are we testing?

src/main/java/dev/dbos/transact/database/StepsDAO.java
```diff
@@ -217,6 +215,58 @@ public List<StepInfo> listWorkflowSteps(String workflowId) throws SQLException {
 
         return steps;
     }
+
+    public double sleep(String workflowUuid, int functionId, double seconds, boolean skipSleep) throws SQLException {
```
kraftp, line 219: There's also some redundant code in the notifications implementation that can be merged with this?

src/test/java/dev/dbos/transact/execution/DBOSExecutorTest.java
manojdbos: This is testing that during recovery any remaining sleep time is honored.

I think this is a good way for the unit test. It is much simple that setting a high sleep time in the workflow and then crashing the app

End to end / chaos can do the testing by crashed the app

src/main/java/dev/dbos/transact/database/StepsDAO.java
manojdbos: that was commented out. Now removed.

PR #16, opened 2025-07-22
Workflow timeouts

manojdbos:


src/test/java/dev/dbos/transact/workflow/TimeoutTest.java
```diff
@@ -0,0 +1,310 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import javax.sql.DataSource;
+import java.sql.*;
+import java.time.Instant;
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.*;
```
kraftp, line 24: We also need some tests for how timeouts interact with child workflows. We should verify that deadlines propagate to children and also that timeouts can be set inside a workflow for children and will override the propagated deadline.

src/test/java/dev/dbos/transact/workflow/TimeoutTest.java
```diff
@@ -0,0 +1,310 @@
...
+        String wfid1 = "wf-128";
+        String result;
+
+        DBOSOptions options = new DBOSOptions.Builder(wfid1).timeout(3).build();
+
+        try (SetDBOSOptions id = new SetDBOSOptions(options)){
+            result = simpleService.workWithString("12345");
+        }
+
+        setDelayEpoch(dataSource, wfid1);
```
kraftp, line 282: I don't think this is needed?

manojdbos: For queues yes Not needed. But needed for recovery. Otherwise what does recovery do when we are past the epoch.

kraftp: That's what we need to test--that the deadline is not changed by recovery, so if you try to recover a workflow that's past its deadline, the workflow will timeout.

manojdbos: That is exactly what the test is testing. I artificially move the deadline back and call executeById to create the situation that recovery may encounter

src/test/java/dev/dbos/transact/workflow/TimeoutTest.java
manojdbos: Added tests

src/test/java/dev/dbos/transact/workflow/SimpleServiceImpl.java
```diff
@@ -125,4 +125,52 @@ public String syncWithQueued() {
...
+        logger.info("Child done sleeping for " + sleepSeconds) ;
+        return input ;
+    }
+    
+    @Workflow(name = "longParent")
+    public String longParent(String input, long sleepSeconds, long timeoutSeconds) throws InterruptedException {
+
+        logger.info("In longParent") ;
+        String workflowId = "childwf" ;
+        DBOSOptions options = new DBOSOptions.Builder(workflowId).async().timeout(timeoutSeconds).build();
```
kraftp, line 163: But this doesn't test deadline propagation if the child has its own independently set deadline. We can (and should) keep these tests to test children with their own deadlines, but we also need to test propagation.

manojdbos: ParentChildTimeOut test show child timeout 1  overriding parent 0

parentTimeoutInheritedByChild test should child 0 inheriting parent timeout 1

PR #17, opened 2025-07-25
Manoj/migration

manojdbos:
Replace flyway with dbos migration

### DIFF
--- a/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
+++ b/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
@@ -52,7 +52,7 @@ public class DBOSExecutor {
 
     public void shutdown() {
         workflowRegistry = null ;
-        executorService.shutdown();
+        executorService.shutdownNow();
         systemDatabase.destroy() ;
     }
 


build.gradle.kts
```diff
@@ -31,8 +28,8 @@ repositories {
 dependencies {
     api("org.slf4j:slf4j-api:2.0.13") // logging api
 
-    implementation("org.flywaydb:flyway-core:10.11.0") // migration
-    implementation("org.flywaydb:flyway-database-postgresql:10.11.0")
+ //   implementation("org.flywaydb:flyway-core:10.11.0") // migration
+ //   implementation("org.flywaydb:flyway-database-postgresql:10.11.0")
```
qianl15, line 32:
```suggestion
```

src/main/java/dev/dbos/transact/migrations/MigrationManager.java
```diff
@@ -0,0 +1,215 @@
...
+import java.io.UncheckedIOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.nio.file.*;
+import java.sql.*;
+import java.util.*;
+import java.util.stream.Collectors;
+
+public class MigrationManager {
```
kraftp, line 23: We can make this MUCH simpler. Since we're doing our own migrations, they don't have to be files--they can just be Java strings containing the SQL. Then we don't need all this file management stuff, we just operate on strings in a big array of migrations.

PR #19, opened 2025-07-29
Fork and restart

manojdbos:


src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -365,5 +365,44 @@ public WorkflowHandle<?> resumeWorkflow(String workflowId) {
...
+    /**
+     * Fork the workflow. Re-execute with another Id from the step provided. Steps prior to the provided
+     * step are copied over
+     *
+     * @param workflowId Original workflow Id
+     * @param startStep Start execution from this step. Prior steps copied over
+     * @param applicationVersion The version of the application to run
+     * @return handle to the workflow
+     */
+    public WorkflowHandle<?> forkWorkflow(String workflowId, int startStep, String applicationVersion) {
```
manojdbos: applicationVersion is not yet implemented.
Needs to be implemented across the board

qianl15: We can make it easy to set app version through DBOS config/launch.

src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
```diff
@@ -494,4 +513,19 @@ public void cancelWorkflow(String workflowId)  {
 
     }
 
+    public WorkflowHandle<?> forkWorkflow(String workflowId, int startStep, String applicationVersion) {
+
+        String newId = UUID.randomUUID().toString();
```
qianl15: I think fork should support user specifying a new ID. In Python and TS, we check if the user provided a new ID first; if not, we generate a new one.

src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -365,5 +365,44 @@ public WorkflowHandle<?> resumeWorkflow(String workflowId) {
...
+    }
+
+    /**
+     * Rerun this workflow from the start with another generated workflow Id
+     *
+     * @param workflowId Original workflow id
+     * @return handle to the workflow
+     */
+
+    public WorkflowHandle<?> restartWorkflow(String workflowId) {
```
qianl15: We plan to deprecate restart across languages because they should all be replaced by fork. I think we can remove it here too.

src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -365,5 +365,48 @@ public WorkflowHandle<?> resumeWorkflow(String workflowId) {
...
+     * Fork the workflow. Re-execute with another Id from the step provided. Steps prior to the provided
+     * step are copied over
+     *
+     * @param workflowId Original workflow Id
+     * @param forkedWorkflowId Id of the new workflow
+     * @param startStep Start execution from this step. Prior steps copied over
+     * @param applicationVersion The version of the application to run
+     * @return handle to the workflow
+     */
+    public WorkflowHandle<?> forkWorkflow(String workflowId, String forkedWorkflowId, int startStep, String applicationVersion) {
```
kraftp: Shouldn't `workflowId` and `startStep` be the first two arguments for consistency?

PR #20, opened 2025-07-30
AdminAPI

manojdbos:


src/main/java/dev/dbos/transact/execution/RecoveryService.java
```diff
@@ -51,4 +56,109 @@ public List<GetPendingWorkflowsOutput> getPendingWorkflows() throws SQLException
...
+     * Background thread method that attempts to recover local pending workflows on startup.
+     * This method runs continuously until stop is requested or all workflows are recovered.
+     */
+    private void startupRecoveryThread(List<GetPendingWorkflowsOutput> wToRecover) {
+        try {
+            List<GetPendingWorkflowsOutput> pendingWorkflows = new CopyOnWriteArrayList<>(wToRecover);
+
+            logger.info("Starting recovery thread " + pendingWorkflows.size()) ;
+
+            while (!stopRequested && !pendingWorkflows.isEmpty()) {
```
kraftp, line 117: I know we have this logic in Python, but I don't think we need it in Java--if I understand correctly, all workflows are required to be registered before launch, so we can just try to recover them, we don't need this loop.

manojdbos: If I understand your comment  -- you are saying dont worry about the outer loop and removing the entry stuff from pendingWorkflows.

You are saying this is enough


for (GetPendingWorkflowsOutput pendingWorkflow : currentPending) {
                        if (stopRequested) {
                            break;
                        }
                        recoverWorkflow(pendingWorkflow);
                     
                    }

PR #25, opened 2025-08-05
StartWorkflow/EnqueueWorkflow

manojdbos:


src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -273,7 +272,7 @@ public void shutdown() {
         }
     }
 
-    public static WorkflowHandle<?> retrieveWorkflow(String workflowId) {
+    public static <R> WorkflowHandle<R> retrieveWorkflow(String workflowId) {
```
devhawk: Is there a reason why we didn't use <T> here like other methods below?

src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
```diff
@@ -435,7 +432,7 @@ public <T> T runStep(String stepName,
      * Retrieve the workflowHandle for the workflowId
      *
      */
-    public WorkflowHandle<?> retrieveWorkflow(String workflowId) {
+    public <R> WorkflowHandle<R> retrieveWorkflow(String workflowId) {
```
devhawk, line 435: Same question from DBOS.java - why <R> instead of <T>?

manojdbos: Making it more strongly typed

src/test/java/dev/dbos/transact/queue/QueuesTest.java
```diff
@@ -178,30 +179,33 @@ public void multipleQueues() throws Exception{
...
-                .queue(secondQ)
                 .build() ;
 
 
         String id1 = "firstQ1234" ;
         String id2 = "second1234" ;
 
-        try (SetWorkflowID ctx = new SetWorkflowID(id1)) {
-            serviceQ1.simpleQWorkflow("firstinput");
+        DBOSOptions options1 = new DBOSOptions.Builder(id1).queue(firstQ).build();
```
devhawk, line 188: Q: it looks like this change moves the enqueueing from the workflow to the callsite. In other words, in the original call when queue was set for the workflow via the WorkflowBuilder.queue method *every* call to that workflow went thru the queue. Now, we can invoke the service directly, but we're using the options at the workflow invocation to specify the queue. Is that correct?

manojdbos: In earlier impl proxy could do only one type of invocation.  Newer proxy can examine context and do all types of invocation. SetDBOSOptions manipulates the context

PR #28, opened 2025-08-06
Update Admin Controller

devhawk:
Implements most of the admin controller methods with tests. 

Unimplemented ones marked with TODO for later implementation

### DIFF
--- a/src/main/java/dev/dbos/transact/DBOS.java
+++ b/src/main/java/dev/dbos/transact/DBOS.java
@@ -211,6 +211,7 @@ public class DBOS {
 
         if (notificationService == null) {
             notificationService = systemDatabase.getNotificationService();
+            notificationService.setInternalWorkflowsService(dbosExecutor.createInternalWorkflowsService(this));
             notificationService.start();
         } else {
             notificationService.start();
--- a/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
+++ b/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
@@ -3,6 +3,7 @@ package dev.dbos.transact.execution;
 import static dev.dbos.transact.exceptions.ErrorCode.UNEXPECTED;
 
 import dev.dbos.transact.Constants;
+import dev.dbos.transact.DBOS;
 import dev.dbos.transact.config.DBOSConfig;
 import dev.dbos.transact.context.DBOSContext;
 import dev.dbos.transact.context.DBOSContextHolder;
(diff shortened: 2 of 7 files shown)


src/main/java/dev/dbos/transact/http/controllers/AdminController.java
```diff
@@ -27,79 +32,200 @@ public AdminController(SystemDatabase s, DBOSExecutor e) {
...
+    public List<String> recovery(List<String> executorIds) {
+        // TODO: implement dbosExec.recoverPendingWorkflows
+        throw new WebApplicationException("Not implemented", Response.Status.INTERNAL_SERVER_ERROR);
+    }
+
+    @POST
+    @Path("/dbos-garbage-collect")
+    public List<String> garbageCollect() {
+        // TODO: implement systemDatabase.garbageCollect
+        throw new WebApplicationException("Not implemented", Response.Status.INTERNAL_SERVER_ERROR);
```
manojdbos: Is this our class ? Not seeing a file or import

src/main/java/dev/dbos/transact/http/controllers/AdminController.java
```diff
@@ -27,79 +32,200 @@ public AdminController(SystemDatabase s, DBOSExecutor e) {
...
     @Path("/workflows/{workflowId}/cancel")
-    @Produces(MediaType.APPLICATION_JSON)
-    public void cancel(@PathParam("workflowId") String workflowId) {
+    public Response cancel(@PathParam("workflowId") String workflowId) {
+        logger.info("Cancel workflow {}", workflowId);
+        dbosExecutor.cancelWorkflow(workflowId);
+        return Response.noContent().build();
+    }
+
+    public static class ForkWorkflowRequest {
```
manojdbos, line 185: Static class might be ok for ForkWorklowRequest and Response.

But the other 2 classes might be reusable when I expose the same API in dbos.

I prefer to use public static sparingly

src/main/java/dev/dbos/transact/http/controllers/AdminController.java
```diff
@@ -27,79 +32,200 @@ public AdminController(SystemDatabase s, DBOSExecutor e) {
...
-    @Path("/recovery")
+    @Path("/queues")
+    @Consumes(MediaType.APPLICATION_JSON)
     @Produces(MediaType.APPLICATION_JSON)
-    public List<String> recovery() {
-
-        return new ArrayList<>();
+    public List<WorkflowStatus> listQueuedWorkflow() {
+        // TODO: implement dbosExec.listQueuedWorkflows
+        throw new WebApplicationException("Not Implemented", Response.Status.INTERNAL_SERVER_ERROR);
```
manojdbos: Can the Response Object be made a parameter so you can set the return status code there -- rather than throwing an exception ? I am thinking something like that would mean you have more control that letting the framework handle the exception

src/main/java/dev/dbos/transact/http/controllers/AdminController.java
devhawk: It's in the jakarta.ws.rs package, which this file imports via wildcard

src/main/java/dev/dbos/transact/http/controllers/AdminController.java
devhawk: Sure, I can modify the not implemented endpoints to return an error code via Response object instead of throwing

PR #29, opened 2025-08-07
Various changes

manojdbos:
Step outside workflow
Global params : App version , executor Id
DBOS.listworflows, listworkflowsteps, listqueuedworkflows
rename DBOSOptions to WorkflowOptions
updated readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,24 +38,34 @@ You add durable workflows to your existing Java program by annotating ordinary f
 
 ```java
 
-public interface SimpleWorkflowService {
-
-    void setSimpleWorkflowService(SimpleWorkflowService e);
+public interface WorkflowService {
+    void setStepService(StepService stepService);
     String exampleWorkflow(String input) ;
+}
+
+public interface StepService {
     void stepOne() ;
     void stepTwo() ;
 }
 
-public class SimpleWorkflowServiceImpl implements SimpleWorkflowService {
+public class WorkflowServiceImpl implements WorkflowService {
 
-    public void setSimpleWorkflowService(SimpleWorkflowService simpleWorkflow) {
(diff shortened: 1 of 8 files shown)


src/main/java/dev/dbos/transact/utils/GlobalParams.java
```diff
@@ -0,0 +1,155 @@
...
+
+    private String appVersion;
+    private String executorId;
+
+    private DBOSExecutor dbosExecutor;
+    private static GlobalParams instance;
+
+    private GlobalParams(DBOSExecutor de) {
+        this.dbosExecutor = de;
+        appVersion = System.getenv("DBOS__APPVERSION") == null
```
kraftp, line 25: You should also be able to set application version programmatically during DBOS configuration, like here: https://docs.dbos.dev/python/reference/configuration

src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
```diff
@@ -332,8 +336,11 @@ public <T> T runStep(String stepName, boolean retriedAllowed, int maxAttempts,
         String workflowId = ctx.getWorkflowId();
 
         if (workflowId == null) {
-            throw new DBOSException(UNEXPECTED.getCode(),
-                    "No workflow id. Step must be called from workflow");
+            // throw new DBOSException(UNEXPECTED.getCode(),
+            // "No workflow id. Step must be called from workflow");
+            logger.warn("Step executed outside a workflow. DBOS features like Checkpointing " +
```
kraftp, line 341: I wouldn't even warn, just call the step normally. The common pattern is that code for a step (which is just a regular function) is shared between a DBOS workflow and non-DBOS code, and it's totally fine for the non-DBOS code to call that step.

PR #30, opened 2025-08-08
Conductor support

devhawk:


src/main/java/dev/dbos/transact/conductor/Conductor.java
```diff
@@ -0,0 +1,436 @@
...
+                                    });
+                        }
+                    }).join();
+        } catch (Exception e) {
+            logger.warn("Error in conductor loop. Reconnecting", e);
+            resetWebSocket();
+        }
+    }
+
+    BaseResponse getResponse(BaseMessage message) {
```
manojdbos, line 332: If you have time .The idiomatic java way of writing (instead of the long switch) this could be:

Define a message handler interface with say a handleMessage method.
Have various impls for each message type.
Keep a map key = messageType, value = handler
To handle a message
look up the handler from map and call handleMessage on it.

src/test/java/dev/dbos/transact/conductor/ConductorTests.java
```diff
@@ -0,0 +1,375 @@
...
+    void beforeEach() throws Exception {
+        testServer = new TestWebSocketServer(0);
+        testServer.start();
+        testServer.waitStart(1000);
+
+        int port = testServer.getPort();
+        assertTrue(port != 0, "Invalid Web Socket Server port");
+        String domain = String.format("ws://localhost:%d", port);
+
+        mockDB = mock(SystemDatabase.class);
```
manojdbos, line 74: All our others tests call SystemDatabase and use a real postgres.
Same for DBOSExecutor.
So you really need to mock them ?

Might be ok if you just testing the websocket classes and not the entire API

devhawk: Much easier to setup mocks than to setup the db for tests

build.gradle.kts
```diff
@@ -54,6 +54,7 @@ dependencies {
     testImplementation("io.rest-assured:xml-path:5.4.0")
     testImplementation(platform("org.junit:junit-bom:5.10.0"))
     testImplementation("org.junit.jupiter:junit-jupiter")
+    testImplementation("org.java-websocket:Java-WebSocket:1.5.6")
```
manojdbos, line 57: Why is this dependency in test only ?

what websockets are we using for runtime ?

manojdbos: For runtime I think u are using java 11 websockets

devhawk: yes, I'm using java.net.http.WebSocket for the websocket client in the production code. Java-WebSocket is used for building a websocket server in the tests

PR #32, opened 2025-08-12
Executor Info Conductor Support

devhawk:


src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
```diff
@@ -59,6 +59,14 @@ public String getAppName() {
         return config.getName();
     }
 
+    public String getExecutorId() {
```
devhawk, line 62: Unfortunetly, the GlobalParams statics can not be mocked effectively, making it nearly impossible to isolation test any component that depend on it. 

I'm not a fan of any global state based design. I don't like the global params class in TS and was working on removing it when I moved over to work on Java. I think it's a bad idea to have them in transact-java, but this workaround solves the issue in the short term until a better solution can be implemented

manojdbos, line 62: Can you think about how to change the GlobalParams class a little bit to help you write the test ?

Add methods here so you can write a test easily seems odd

src/test/java/dev/dbos/transact/conductor/ConductorTests.java
```diff
@@ -237,6 +240,36 @@ public void send(BaseMessage message) throws Exception {
...
     }
 
+    @Test
+    public void canExecutorInfo() throws Exception {
+        MessageListener listener = new MessageListener();
+        testServer.setListener(listener);
+
+        String hostname = InetAddress.getLocalHost().getHostName();
+
+        when(mockExec.getAppVersion()).thenReturn("test-app-version");
```
manojdbos, line 250: What if you added a constructor to globalparams ? Would that not do the job . You would not event need to mock then

devhawk: We would need a global params public ctor and a public setter for the global instance. Mocking is much simpler than that

src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
devhawk: The GlobalParams class is a classic example of the [Global Data code smell](https://luzkan.github.io/smells/global-data). I'd like to get away from that design (in a future PR)

src/test/java/dev/dbos/transact/conductor/ConductorTests.java
manojdbos: In general so far we have stayed away from mocking.

I dont recall mocking much in TS or Python as well.

devhawk: TS tests take 20 minutes to run and Python takes 15. Not sure that's a model worth emulating.

manojdbos: You can also pass in the values using config.

src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
manojdbos: This is how it is done in other languages.

The java project is not at the stage for redesign.

Also we should not add methods to DBOSExecutor just because it is easier to write the test.

manojdbos: I have seen the above comment at large companies like IBM or Microsoft where they have lots of time and resource.

I find it hard to believe we are talking about this  especially after what Jeremy said in the company meeting.

I would recommend you fix it in other languages first. If time permits , I will fix it here

PR #35, opened 2025-08-14
Fixes for the cloud

manojdbos:


src/main/java/dev/dbos/transact/migrations/MigrationManager.java
```diff
@@ -63,7 +63,7 @@ public static void runMigrations(DBOSConfig dbconfig) {
     }
 
     public static void createDatabaseIfNotExists(DBOSConfig config, String dbName) {
-        DataSource adminDS = SystemDatabase.createDataSource(config, Constants.POSTGRES_DEFAULT_DB);
+        DataSource adminDS = SystemDatabase.createPortgresDataSource(config);
```
chuck-dbos: Postgres?

src/main/java/dev/dbos/transact/database/SystemDatabase.java
```diff
@@ -462,6 +462,26 @@ public static DataSource createDataSource(DBOSConfig config, String dbName) {
...
+                config.getDbPort(),
+                Constants.POSTGRES_DEFAULT_DB);
+
+        String dbUser = config.getDbUser();
+        String dbPassword = config.getDbPassword();
+        hikariConfig.setJdbcUrl(dburl);
+        hikariConfig.setUsername(dbUser);
+        hikariConfig.setPassword(dbPassword);
+
+        hikariConfig.setMaximumPoolSize(2);
```
qianl15, line 481: The pool size maybe need to be larger because it limits the number of concurrent operations -- it's set to 20 for both TS and Python.

manojdbos: It is a short lived one time connection to the postgres db to create the systemdb if it is not there.

PR #43, opened 2025-08-18
Publish to GH Packages

devhawk:
Also includes a basic versioning scheme that will need to be revisited before shipping the next minor version.

# Versioning Scheme

The `build.gradle.kts` file includes the following logic for generating the version identifier:

`version = "$baseVersion.$commitCount-preview+g$gitHash$safeBranchName"`

* `baseVersion` is set to `BASE_VERSION` env var, with `0.5` as a fallback
* `commitCount` is the number of commits, retrieved via `git rev-list --count HEAD`
* `gitHash` is the short git hash, retrieved via ` git rev-parse --short HEAD`
* `safeBranchName` is the branch name, with `/` replaced with `-`. Omitted if branch name is `main` or `HEAD`

An example version identifier from my local machine is `0.5.302-preview+ga068c39.devhawk-version`.

### DIFF
--- a/src/test/java/dev/dbos/transact/queue/QueuesTest.java
+++ b/src/test/java/dev/dbos/transact/queue/QueuesTest.java
@@ -486,7 +486,6 @@ public class QueuesTest {
     }
 
     @Test
-    @Disabled
     public void testQueueConcurrencyUnderRecovery() throws Exception {
 
         Queue queue = new DBOS.QueueBuilder("test_queue").concurrency(2).build();


.github/workflows/publish.yml
```diff
@@ -0,0 +1,33 @@
+name: Publish
+
+on:
+  workflow_call:
+   secrets:
+    USERNAME_MAVEN:
+      required: true
+    TOKEN_MAVEN:
+      required: true
```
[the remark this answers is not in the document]
devhawk: I accepted this on my phone w/o being able to see the full diff. THis is a poor suggestion and I've reverted

build.gradle.kts
```diff
@@ -1,3 +1,51 @@
...
+            }
+        }.toString().trim()
+    }
+}
+
+// Note, this versioning scheme is fine for preview releases
+// but we'll want something more robust once we want to bump
+// the major or minor version number
+val baseVersion = System.getenv("BASE_VERSION") ?: "0.5"
+val safeBranchName = if (branchName == "main" || branchName == "HEAD") "" else ".${branchName.replace("/", "-")}"
```
kraftp, line 44: Fine for now, but we'll probably need to add a tag-based system like Python/TS before release.

devhawk: absolutely. Hell, we'll probably want to revisit before we ship 0.6

PR #47, opened 2025-08-19
Add conductorDomain to DBOSConfig

devhawk:
~* move conductor key from config to dbos launch param~
~* Add conductor domain to dbos launch param~

As per PR feedback, changed to leave conductor key where it was and just add conductor domain to dbos config.


src/main/java/dev/dbos/transact/config/DBOSConfig.java
```diff
@@ -57,7 +55,6 @@ public static class Builder {
         private int httpPort;
```
kraftp: It looks like this file could be cleaned up--a bunch of these options aren't used anymore

devhawk: I don't want to get into general cleanup in this PR w/o @manojdbos input

Issue #51, opened 2025-08-21
calling dbos.launch after dbos.shutdown throws null pointer exception


devhawk:
```
java.lang.NullPointerException: Cannot invoke "dev.dbos.transact.execution.WorkflowRegistry.getClasses()" because "this.workflowRegistry" is null
        at dev.dbos.transact.execution.DBOSExecutor.getRegisteredClasses(DBOSExecutor.java:634)
        at dev.dbos.transact.execution.DBOSExecutor.start(DBOSExecutor.java:93)
        at dev.dbos.transact.DBOS.launch(DBOS.java:210)
        at dev.dbos.demo.ConductorApp.recreateEndpoint(ConductorApp.java:83)
        at spark.RouteImpl$1.handle(RouteImpl.java:72)
        at spark.http.matching.Routes.execute(Routes.java:61)
        at spark.http.matching.MatcherFilter.doFilter(MatcherFilter.java:134)
        at spark.embeddedserver.jetty.JettyHandler.doHandle(JettyHandler.java:50)
        at org.eclipse.jetty.server.session.SessionHandler.doScope(SessionHandler.java:1598)
        at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:141)
        at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:127)
        at org.eclipse.jetty.server.Server.handle(Server.java:516)
        at org.eclipse.jetty.server.HttpChannel.lambda$handle$1(HttpChannel.java:487)
        at org.eclipse.jetty.server.HttpChannel.dispatch(HttpChannel.java:732)
        at org.eclipse.jetty.server.HttpChannel.handle(HttpChannel.java:479)
        at org.eclipse.jetty.server.HttpConnection.onFillable(HttpConnection.java:277)
        at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:311)
        at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:105)
        at org.eclipse.jetty.io.ChannelEndPoint$1.run(ChannelEndPoint.java:104)
        at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:883)
        at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:1034)
        at java.base/java.lang.Thread.run(Thread.java:840)
```

[...]

PR #52, opened 2025-08-21
Disable code in shutdown path

devhawk:
conductor repo tests need to be able to shutdown and relaunch DBOS. However, the current DBOS shutdown path closes the DB, shuts down the executor service and discards teh worklow registry. 

This PR comments out that code to allow shutdown/relaunch to succeed. Clearly, this is not a long term solution. Issue #51 tracks refactoring the core objects and their dependencies to solve this issue.

### DIFF
--- a/src/test/java/dev/dbos/transact/migration/MigrationManagerTest.java
+++ b/src/test/java/dev/dbos/transact/migration/MigrationManagerTest.java
@@ -31,7 +31,7 @@ class MigrationManagerTest {
     static void setup() throws Exception {
 
         MigrationManagerTest.dbosConfig = new DBOSConfig.Builder().name("migrationtest")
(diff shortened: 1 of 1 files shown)


src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -285,7 +285,7 @@ public void shutdown() {
                 httpServer.stop();
             }
 
-            systemDatabase.destroy();
+            // systemDatabase.destroy();
```
qianl15, line 289: Let's put a TODO notes here just in case we forget about it?

devhawk: Will do though #51 is tracking the real fix

PR #62, opened 2025-08-27
update readme

devhawk:


kraftp (approved):
This should also prominently say that Java is pre-release and under construction

chuck-dbos:
I see links to python docs, that may be intentional until there are java docs?

devhawk:
> I see links to python docs, that may be intentional until there are java docs?

There are no java docs yet, so I left those in

PR #64, opened 2025-08-28
Move WF/Q Registry into DBOS 

### LINKED ISSUE #36: Workflow Registration should not be allowed after DBOS.launch has been called

devhawk:
This PR moves the workflow and queue registry into the main DBOS instance and centralizes all registration work there as well. DBOS passes a snapshot of the registry on launch so that DBOS executor and queue service are always working with an unchanging snapshot.

fixes #36


src/test/java/dev/dbos/transact/workflow/SimpleServiceImpl.java
```diff
@@ -108,21 +107,21 @@ public String grandParent(String input) throws Exception {
     @Workflow(name = "syncWithQueued")
     public String syncWithQueued() {
 
-        System.out.println("In syncWithQueued " + DBOSContext.workflowId().get());
+        throw new IllegalStateException("this test method is broken");
```
kraftp, line 110: Can we clean these bad tests up?

kraftp (approved):
This looks good!

src/test/java/dev/dbos/transact/workflow/SimpleServiceImpl.java
devhawk: I have one more core PR to.do, and I am planning a pass thru the tests to re-enable any tests that I've had to disable. Cool?

PR #66, opened 2025-09-02
Executor Refactor

devhawk:
This PR moves all launch services (recovery, queue, scheduler, sysdb, admin server and conductor) into DBOSExecutor. DBOS Executor instance tied to launch/shutdown (i.e. constructed + started on launch, closed on shutdown)


kraftp (approved):
This looks directionally right. The launch services' lifetime should be tied to the executor.

PR #69, opened 2025-09-04
DBOSClient

devhawk:


src/main/java/dev/dbos/transact/DBOSClient.java
```diff
@@ -0,0 +1,159 @@
...
+
+    public <T> WorkflowHandle<T> retrieveWorkflow(String workflowId) {
+        return new WorkflowHandleDBPoll<T>(workflowId, systemDatabase);
+    }
+
+    public void cancelWorkflow(String workflowId) {
+        systemDatabase.cancelWorkflow(workflowId);
+    }
+
+    public void resumeWorkflow(String workflowId) {
```
kraftp: This and fork should return handles to match the library UX (easy enough by calling retrieve)

src/main/java/dev/dbos/transact/database/SystemDatabase.java
```diff
@@ -422,8 +426,20 @@ private <T> T handleExistingResult(StepResult result, String functionName) {
         }
     }
 
-    public static HikariDataSource createDataSource(DBOSConfig config, String dbName) {
+    public static HikariDataSource createDataSource(String url, String user, String password, int poolSize, int timeout) {
```
kraftp, line 429: Why specify user and password separately from url?

src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
```diff
@@ -80,59 +80,6 @@ public DBOSExecutor(DBOSConfig config) {
         this.config = config;
```
kraftp, line 80: What are we changing in this file?

src/main/java/dev/dbos/transact/database/SystemDatabase.java
devhawk: that's how JDBC works. username and password are not passed in their normal locations in the URL. You can pass them in the URL but as [query params](https://jdbc.postgresql.org/documentation/use/#connection-parameters).

We may want to add a compat layer so that you specify PG URLs as we do in DBOS Cloud and then transform them into the equivalent JDBC URL, but I think that should be a separate PR

src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
devhawk: While I did move stuff around, the primary change was to expose a static `enqueueWorkflow` method that `DBOSClient` can call w/o creating an executor instance.

src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
```diff
@@ -502,30 +537,59 @@ public <T> WorkflowHandle<T> submitWorkflow(String workflowName, String targetCl
...
 
-    public void enqueueWorkflow(String workflowName, String targetClassName,
-            WorkflowFunctionWrapper wrapper, Object[] args, Queue queue) throws Throwable {
-
-        DBOSContext ctx = DBOSContextHolder.get();
-        ctx.setDbos(dbos);
-        String wfid = ctx.getWorkflowId();
+    // TODO: add priority + deduplicationId support
+    // (https://github.com/dbos-inc/dbos-transact-java/issues/67)
+    public static String enqueueWorkflow(SystemDatabase systemDatabase, String wfid, String workflowName,
```
devhawk, line 531: @kraftp this (along with making pre/postInvokeWF static) is the change I mentioned in my [response above](https://github.com/dbos-inc/dbos-transact-java/pull/69#discussion_r2323519662)

kraftp (approved):

src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
```diff
@@ -303,32 +322,74 @@ public WorkflowInitResult preInvokeWorkflow(String workflowName, String classNam
...
+    /**
+     * Run and postInvoke reused separately preInvoke in reused separately
+     *
+     * @param target
+     * @param args
+     * @param function
+     * @param workflowId
+     * @return
+     * @param <T>
+     * @throws Throwable
```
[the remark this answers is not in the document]
devhawk: This is a private API, doesn't need extensive comments

PR #78, opened 2025-09-11
Chuck/runstep

chuck-dbos:


src/main/java/dev/dbos/transact/workflow/StepInterfaces.java
```diff
@@ -0,0 +1,23 @@
...
+    void run() throws E;
+  }
+
+  @FunctionalInterface
+  public interface NonthrowingNoArg<R> {
+    R get();
+  }
+
+  @FunctionalInterface
+  public interface NonthrowingNoArgNoReturn {
```
devhawk: we just use 'java.lang.Runnable` instead of defining `NonthrowingNoArgNoReturn`

src/main/java/dev/dbos/transact/workflow/StepInterfaces.java
```diff
@@ -0,0 +1,23 @@
...
+    R get() throws E;
+  }
+
+  @FunctionalInterface
+  public interface ThrowingNoArgNoReturn<E extends Exception> {
+    void run() throws E;
+  }
+
+  @FunctionalInterface
+  public interface NonthrowingNoArg<R> {
```
devhawk: use `java.util.function.Supplier<T>`

src/main/java/dev/dbos/transact/workflow/StepInterfaces.java
```diff
@@ -0,0 +1,23 @@
+package dev.dbos.transact.workflow;
+
+public class StepInterfaces {
+  @FunctionalInterface
+  public interface ThrowingNoArg<R, E extends Exception> {
```
devhawk: Can we use the already defined `ThrowingRunnable` and `ThrowingSupplier<T>` ?

src/main/java/dev/dbos/transact/workflow/StepInterfaces.java
chuck-dbos: It was unused.

src/main/java/dev/dbos/transact/workflow/StepOptions.java
```diff
@@ -0,0 +1,75 @@
+package dev.dbos.transact.workflow;
+
+public record StepOptions(
+    String name,
+    boolean retriesAllowed,
+    int maxAttempts,
+    double intervalSeconds,
+    double backOffRate) {
+  public static double DEFAULT_INTERVAL_SECONDS = 1.0;
```
devhawk: Can we see if we can reuse these constants in the Step annotation class?

devhawk (approved):
minor nit about using step option default constants in Step annotation. Otherwise looks ready to merge

src/main/java/dev/dbos/transact/workflow/StepOptions.java
chuck-dbos: Yes that makes good sense.

PR #83, opened 2025-09-15
Implement step retry + tests for same

chuck-dbos:


src/main/java/dev/dbos/transact/workflow/Step.java
```diff
@@ -12,9 +12,9 @@
 
   boolean retriesAllowed() default false;
 
-  float intervalSeconds() default 1.0f;
+  double intervalSeconds() default 1.0f;
```
devhawk: We should change defaults from 1.0f to 1.0

src/test/java/dev/dbos/transact/step/StepsTest.java
```diff
@@ -235,4 +235,44 @@ public void stepOutsideWorkflow() throws Exception {
     result = serviceB.step2("pqrstu");
     assertEquals("pqrstu", result);
   }
+
+  @Test
+  public void StepRetryLogic() throws Exception {
```
devhawk: nit: method should start with lower case

src/test/java/dev/dbos/transact/step/ServiceWFAndStepImpl.java
```diff
@@ -28,4 +28,103 @@ public String stepOne(String input) {
...
     return input;
   }
+
+  int stepWithRetryRuns = 0;
+
+  @Step(
+      name = "stepWith2Retries",
+      retriesAllowed = true,
+      maxAttempts = 2,
+      intervalSeconds = .01f,
```
devhawk: we changed to double, right?

src/test/java/dev/dbos/transact/step/StepsTest.java
chuck-dbos: In my defense, I followed an example from above.  Making them all consistent...

src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
```diff
@@ -725,7 +725,8 @@ public <T> T runStep(
       String stepName,
       boolean retriedAllowed,
       int maxAttempts,
-      float backOffRate,
+      double timeBetweenAttempts,
```
devhawk: I'm trying to move code over to use Duration type. I'm OK using double here since we're simply passing thru the values from annotation (which have to be primitives) but we should indicate time here is in seconds.

chuck-dbos: Yep will change name for now, would like to use the options struct once we merge the next branch in though?

devhawk: yes, I think using step options from the runStep branch would be a good idea here. Will need to add retry interval to that type

devhawk (approved):

PR #84, opened 2025-09-17
Fix step and workflow exception returns

chuck-dbos:
Steps and Workflows now throw their original errors, rather than a variety of RuntimException wrappers.  These exceptions are now serialized to the database so that they can be caught by type, as is the norm in Java.

A smattering of test cleanups as well.

### DIFF
--- a/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
+++ b/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
@@ -708,7 +708,7 @@ public class DBOSExecutor implements AutoCloseable {
   // TODO: should these also throw DBOS exceptions?
   // Should there be an unchecked version that promotes errors to unchecked?
   @SuppressWarnings("unchecked")
-  public <R, E extends Exception> R runStepI(ThrowingSupplier<R, E> stepfunc, StepOptions opts)
+  public <T, E extends Exception> T runStepI(ThrowingSupplier<T, E> stepfunc, StepOptions opts)
       throws E {
     try {
       return runStepInternal(
@@ -845,8 +845,8 @@ public class DBOSExecutor implements AutoCloseable {
   }
 
   /** Retrieve the workflowHandle for the workflowId */
-  public <R, E extends Exception> WorkflowHandle<R, E> retrieveWorkflow(String workflowId) {
-    return new WorkflowHandleDBPoll<R, E>(workflowId, systemDatabase);
+  public <T, E extends Exception> WorkflowHandle<T, E> retrieveWorkflow(String workflowId) {
+    return new WorkflowHandleDBPoll<T, E>(workflowId, systemDatabase);
   }
 
   @SuppressWarnings("unchecked")
--- a/src/main/java/dev/dbos/transact/DBOS.java
+++ b/src/main/java/dev/dbos/transact/DBOS.java
@@ -389,7 +389,7 @@ public class DBOS {
     executor.sleep(seconds);
   }
 
-  public <R, E extends Exception> R runStep(ThrowingSupplier<R, E> stepfunc, StepOptions opts)
+  public <T, E extends Exception> T runStep(ThrowingSupplier<T, E> stepfunc, StepOptions opts)
       throws E {
     var executor = dbosExecutor.get();
     if (executor == null) {


src/main/java/dev/dbos/transact/database/StepsDAO.java
```diff
@@ -212,8 +212,8 @@ public List<StepInfo> listWorkflowSteps(String workflowId) throws SQLException {
           Exception error = null;
           if (errorData != null) {
             try {
-              // TODO error = JSONUtil.deserialize(errorData);
-              error = new Exception(errorData);
+              // TODO This should not be at this layer in the stack;
```
devhawk, line 215: truth

src/main/java/dev/dbos/transact/json/JSONUtil.java
```diff
@@ -63,4 +77,77 @@ public static <T> T fromJson(String content, Class<T> valueType) {
...
   }
+
+  public static final class WireThrowable {
+    public int v = 1;
+    public String type;
+    public String message;
+    public String causeType;
+    public List<String> stackPreview;
+    public int suppressedCount;
+    public SerializedPayload serialized;
```
devhawk: Why are we using a wrapper class for the payload? I'm sure there's a good reason, just wanted to know what it was.

src/main/java/dev/dbos/transact/json/JSONUtil.java
```diff
@@ -63,4 +77,77 @@ public static <T> T fromJson(String content, Class<T> valueType) {
...
+    public String ts; // ISO-8601
```
devhawk: can we spell out timestamp

src/main/java/dev/dbos/transact/json/JSONUtil.java
chuck-dbos: I will take that out.  The suggestion was that the payload had an encoding, crc, and other crap, but I think a string ought to be plenty.

src/main/java/dev/dbos/transact/json/JSONUtil.java
chuck-dbos: We do not need it at all, we have sufficient other fields in the DB.

src/main/java/dev/dbos/transact/DBOSClient.java
```diff
@@ -80,7 +80,7 @@ public EnqueueOptions build() {
     }
   }
 
-  public <T> WorkflowHandle<T> enqueueWorkflow(EnqueueOptions options, Object[] args)
+  public <T> WorkflowHandle<T, ?> enqueueWorkflow(EnqueueOptions options, Object[] args)
```
chuck-dbos, line 83: I am not sold on this prototype.... or why it would throw `Exception`.

src/main/java/dev/dbos/transact/json/JSONUtil.java
```diff
@@ -37,15 +47,19 @@ public static Object[] deserializeToArray(String json) {
...
 
-  public static SerializableException deserializeError(String json) {
-    Object[] eArray = JSONUtil.deserializeToArray(json);
-    return (SerializableException) eArray[0];
+  public static Throwable deserializeAppException(String str) {
+    var wt = (WireThrowable) deserializeToArray(str)[0];
+    try {
+      return WireThrowableCodec.toThrowable(wt, null);
+    } catch (Exception e) {
+      System.err.println("Didn't deserialize");
```
devhawk: we should be using slf4j logger here, not syserr

src/main/java/dev/dbos/transact/workflow/internal/WorkflowHandleFuture.java
```diff
@@ -27,7 +27,7 @@ public String getWorkflowId() {
   }
 
   @Override
-  public T getResult() {
+  public T getResult() throws E {
```
devhawk, line 28: should we be matching the code in `WorkflowHandleDBPoll` and cast the thrown exception to `E` rather than throwing an unexpected `DBOSException`?

src/main/java/dev/dbos/transact/DBOSClient.java
devhawk: This should have the same method generic as `retrieveWorkflow`.

src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
```diff
@@ -236,7 +236,7 @@ public Optional<Queue> getQueue(String queueName) {
     return Optional.empty();
   }
 
-  WorkflowHandle<?> recoverWorkflow(GetPendingWorkflowsOutput output) throws Exception {
+  WorkflowHandle<?, ?> recoverWorkflow(GetPendingWorkflowsOutput output) throws Exception {
```
devhawk, line 239: Should we be making these methods generic on return and exception instead of unspecified?

src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
```diff
@@ -489,7 +488,7 @@ public <T> T syncWorkflow(
     return runWorkflowAndSaveResult(target, args, function, workflowId);
   }
 
-  public <T> WorkflowHandle<T> submitWorkflow(
+  public <T> WorkflowHandle<T, ?> submitWorkflow(
```
devhawk, line 491: this really feels like we should specify both return and exception types or neither

src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
```diff
@@ -832,22 +835,22 @@ public <T> T runStepInternal(
...
+              workflowId, stepFunctionId, stepName, null, JSONUtil.serializeAppException(eThrown));
       systemDatabase.recordStepResultTxn(stepResult);
       throw eThrown;
     }
   }
 
   /** Retrieve the workflowHandle for the workflowId */
-  public <R> WorkflowHandle<R> retrieveWorkflow(String workflowId) {
-    return new WorkflowHandleDBPoll<R>(workflowId, systemDatabase);
+  public <R, E extends Exception> WorkflowHandle<R, E> retrieveWorkflow(String workflowId) {
```
devhawk: nit, we should be consistent about using `T` or `R`

src/main/java/dev/dbos/transact/workflow/internal/WorkflowHandleFuture.java
chuck-dbos: Yes, I was a bit confused here and whether your branch changes it.  But I will rethrow as E for now.

src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
chuck-dbos: I feel not.  There are a lot of cases where the type either does not matter, or can be inferred, and making it explicit is therefore a pain / gets you a warning.

src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
chuck-dbos: Do you want T or R?

PR #92, opened 2025-09-25
Add tests for WF and step errors

chuck-dbos:


src/main/java/dev/dbos/transact/database/WorkflowDAO.java
```diff
@@ -420,33 +420,33 @@ public List<WorkflowStatus> listWorkflows(ListWorkflowsInput input) throws SQLEx
     // --- WHERE Clauses ---
     StringJoiner whereConditions = new StringJoiner(" AND ");
 
-    if (input.getWorkflowName() != null) {
+    if (input != null && input.getWorkflowName() != null) {
```
devhawk: maybe put all these inside a single input != null block?

src/main/java/dev/dbos/transact/workflow/StepInfo.java
```diff
@@ -40,7 +47,11 @@ public String getChildWorkflowId() {
   public String toString() {
     return String.format(
         "StepInfo{functionId=%d, functionName='%s', output=%s, error=%s, childWorkflowId='%s'}",
-        functionId, functionName, output, error, childWorkflowId);
+        functionId,
+        functionName,
```
devhawk, line 51: we should just make step info a record

chuck-dbos: This PR or another PR?

src/main/java/dev/dbos/transact/workflow/StepInfo.java
```diff
@@ -1,14 +1,21 @@
 package dev.dbos.transact.workflow;
 
 public class StepInfo {
+  public static record StepInfoError(
+      String className, String message, String serializedError, Throwable throwable) {}
```
devhawk, line 5: this is the same as WorkflowStatusError? Mabe we just have an ExecptionInfo record shared between both?

src/main/java/dev/dbos/transact/workflow/WorkflowStatus.java
```diff
@@ -15,8 +18,7 @@ public class WorkflowStatus {
   private String[] authenticatedRoles;
   private Object[] input;
   private Object output;
-  // TODO fix private Throwable error;
-  private String error;
+  private WorkflowStatusError error;
```
devhawk, line 21: we should make workflow status a record to, but probably in a separate PR

devhawk (approved):

devhawk (approved):
minor nits, looks good

src/main/java/dev/dbos/transact/workflow/StepInfo.java
chuck-dbos: It's the same, can't think of a reason they would become different, but Java makes code reuse for something tiny a lot more painful than it should be.

PR #94, opened 2025-09-26
Step info and WF status become records

chuck-dbos:
Prior PR feedback - a lot of churn involved, so made a separate PR.

### DIFF
--- a/src/main/java/dev/dbos/transact/workflow/StepInfo.java
+++ b/src/main/java/dev/dbos/transact/workflow/StepInfo.java
@@ -1,16 +1,8 @@
 package dev.dbos.transact.workflow;
 
 public record StepInfo(
-    int functionId, String functionName, Object output, ErrorResult error, String childWorkflowId) {
-
-  @Override
-  public String toString() {
-    return String.format(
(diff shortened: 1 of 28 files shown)


src/main/java/dev/dbos/transact/workflow/StepInfo.java
```diff
@@ -1,47 +1,7 @@
...
-
-  public StepInfoError getError() {
-    return error;
-  }
-
-  public String getChildWorkflowId() {
-    return childWorkflowId;
-  }
+public record StepInfo(
+    int functionId, String functionName, Object output, ErrorResult error, String childWorkflowId) {
```
devhawk: We can delete toString as well

src/main/java/dev/dbos/transact/workflow/WorkflowStatus.java
```diff
@@ -226,80 +32,12 @@ public Instant getDeadline() {
...
-        executorId,
-        appVersion,
-        workflowTimeoutMs,
-        workflowDeadlineEpochMs,
-        appId,
-        recoveryAttempts);
+    return null;
   }
 
   @Override
```
devhawk: we can delete toString as well

devhawk (approved):
looks good. Let's just delete the new record `toString` methods

PR #101, opened 2025-10-02
Allow multiple instances of same class

### LINKED ISSUE #90: named instance support

chuck-dbos:
fixes #90


src/main/java/dev/dbos/transact/conductor/protocol/ListWorkflowsRequest.java
```diff
@@ -12,6 +12,8 @@ public class ListWorkflowsRequest extends BaseMessage {
   public static class Body {
     public List<String> workflow_uuids;
     public String workflow_name;
+    public String class_name;
```
devhawk: Please check w/ @qianl15 on this. I don't think conductor supports these fields

src/test/java/dev/dbos/transact/invocation/MultiInstTest.java
```diff
@@ -117,6 +162,79 @@ void startWorkflow() throws Exception {
     assertEquals(hhandle.getWorkflowId(), hrow.workflowId());
     assertEquals("stepWorkflow", hrow.name());
     assertEquals("dev.dbos.transact.invocation.HawkServiceImpl", hrow.className());
+    assertEquals("", hrow.instanceName());
```
devhawk, line 163: shouldn't instance name be null if not set?

src/main/java/dev/dbos/transact/internal/WorkflowRegistry.java
```diff
@@ -10,22 +10,25 @@
 public class WorkflowRegistry {
   private final ConcurrentHashMap<String, RegisteredWorkflow> registry = new ConcurrentHashMap<>();
 
-  public static String getFullyQualifiedWFName(String cls, String wf) {
-    return String.format("%s/%s", cls, wf);
+  public static String getFullyQualifiedWFName(String cls, String instanceName, String wf) {
+    return String.format("%s/%s/%s", cls, instanceName, wf);
```
devhawk, line 14: should we be using Objects.requireNonNull to ensure these strings are not null? Well, maybe instanceName should be nullable?

src/main/java/dev/dbos/transact/conductor/protocol/ListWorkflowsRequest.java
qianl15: Ah we don't pass along instance name or class name: https://github.com/dbos-inc/dbos-transact-py/blob/main/dbos/_conductor/protocol.py#L113-L125

src/test/java/dev/dbos/transact/invocation/MultiInstTest.java
chuck-dbos: We use empty name.

devhawk (approved):

PR #102, opened 2025-10-02
Support DBOS PG urls

### LINKED ISSUE #100: add a DBOSConfig.Builder method for dbUrl

devhawk:
* Introduces a new `DatabaeUrl` record type 
  * can be created from a `jdbc:postgres://` or `postgres://` URL
  * handles user & password specified in query params or URI user info
  * generates `jdbc:` URL sutable for creating a HikariDataSource from 
* simplifies `MigrationManager` by virute of the `withDatabase` method, making it very easy to generate a db url for the postgres DB in `createDatabaseIfNotExists`
* includes `withUser` and `withPassword` methods so DBOSClient can override user/password values from provided URL if desired

Fixes #100


src/main/java/dev/dbos/transact/database/DatabaseUrl.java
```diff
@@ -0,0 +1,122 @@
...
+
+  public DatabaseUrl withUser(String user) {
+    return new DatabaseUrl(scheme(), host(), port(), database(), user, password(), query());
+  }
+
+  public DatabaseUrl withPassword(String password) {
+    return new DatabaseUrl(scheme(), host(), port(), database(), username(), password, query());
+  }
+
+  static Map<String, String> parseQuery(String query) {
```
qianl15, line 40: Is there a standard library in Java to parse URLs? This looks fine but I'm worried about weird strings in the wild, like the one with url encoded that may have random `&` in it.

devhawk: According to chatgpt, not in the standard library

kraftp (commented):
We need to clean this up and just use JDBC URLs (plus username and password, provided separately) everywhere (not just in config). This parsing and custom URL formatting shouldn't be necessary.

src/main/java/dev/dbos/transact/database/DatabaseUrl.java
devhawk: Apparently. the PG JDBC driver has a parseURL function. We already have a dependency on that so I'll see if we can reuse it.

devhawk: https://github.com/pgjdbc/pgjdbc/blob/ef6692365886ad772d1810da568e1ed147906aa4/pgjdbc/src/main/java/org/postgresql/Driver.java#L535

devhawk: parseURL in the PG JDBC driver is only for parsing JDBC PG URLs

PR #103, opened 2025-10-03
Exception handling fixes

### LINKED ISSUE #48: Eliminate JSON checked exceptions
The Jackson JSON parsing library uses checked exceptions, which require catching and re-wrapping. This creates dozens of lines of extraneous stack trace in the log.

Alternatives:
* switch to a JSON libarary that doesn't use checked exceptions like Gson or Moshi
* Wrap Jackson's checked `JsonProcessingException` in a custom unchecked exception that doesn't create a new stack trace

### LINKED ISSUE #86: Cleanup all exceptions

chuck-dbos:
No "throws Exception"
Use Java built-in RuntimeException derivatives for programmer errors
Consistent DBOS exception naming
Use DB retry, and if it fails, system database is in bad shape and throw a (consistent) unchecked exception.

fixes #86 
fixes #48


src/main/java/dev/dbos/transact/exceptions/ErrorCode.java
```diff
@@ -10,9 +10,8 @@ public enum ErrorCode {
   WORKFLOW_CANCELLED(7),
   UNEXPECTED_STEP(8),
   WORKFLOW_FUNCTION_NOT_FOUND(9),
-  SLEEP_NOT_IN_WORKFLOW(10),
-  RESUME_WORKFLOW_ERROR(11),
-  FORK_WORKFLOW_ERROR(12);
+  SYSTEM_DATABASE_ACCESS_ERROR(10),
```
devhawk: Does anything ever actually read this error code?

src/main/java/dev/dbos/transact/exceptions/DBOSAwaitedWorkflowCancelledException.java
```diff
@@ -1,9 +1,9 @@
 package dev.dbos.transact.exceptions;
 
-public class AwaitedWorkflowCancelledException extends DBOSException {
+public class DBOSAwaitedWorkflowCancelledException extends DBOSException {
```
devhawk: Why are we keeping this pattern for DBOS exceptions? (inheriting from DBOSException, including an error code even though the exception is strongly typed). This doesn't seem typical for Java codebases

src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
```diff
@@ -39,7 +39,6 @@
 import dev.dbos.transact.workflow.internal.WorkflowStatusInternal;
 
 import java.lang.reflect.InvocationTargetException;
-import java.sql.SQLException;
 import java.time.Duration;
```
devhawk, line 41: 🎉🎉🎉

devhawk (approved):
Other than the feedback on getting rid of `ErrorCode` and flattening the exception hierarchy, looks good

src/main/java/dev/dbos/transact/exceptions/DBOSAwaitedWorkflowCancelledException.java
chuck-dbos: Happy to try removing it... was certainly borrowed + we weren't all that terribly consistent with it.

src/main/java/dev/dbos/transact/json/JSONUtil.java
```diff
@@ -39,7 +49,7 @@ public static String serializeArray(Object[] args) {
     try {
       return mapper.writeValueAsString(new Boxed(args));
     } catch (JsonProcessingException e) {
-      throw new RuntimeException("Serialization failed", e);
+      throw new JsonRuntimeException(e);
```
devhawk, line 52: 🙏

src/main/java/dev/dbos/transact/database/SystemDatabase.java
```diff
@@ -117,69 +120,83 @@ public void recordWorkflowOutput(String workflowId, String result) {
...
-
-    try {
-      try (Connection connection = dataSource.getConnection()) {
-        return StepsDAO.checkStepExecutionTxn(workflowId, functionId, functionName, connection);
-      }
-    } catch (SQLException sq) {
-      logger.error("Unexpected SQL exception", sq);
-      throw new DBOSException(UNEXPECTED.getCode(), sq.getMessage());
-    }
+      throws IllegalStateException, DBOSWorkflowCancelledException, DBOSUnexpectedStepException {
```
devhawk: Why are we declaring these? Their all runtime exceptions, right?

src/main/java/dev/dbos/transact/database/SystemDatabase.java
```diff
@@ -107,7 +107,10 @@ public InsertWorkflowResult insertWorkflowStatus(
    * @param result output serialized as json
    */
   public void recordWorkflowOutput(String workflowId, String result) {
-    workflowDAO.recordWorkflowOutput(workflowId, result);
+    DbRetry.run(
```
devhawk, line 110: Do we need DbRetry on insertWorkflowStatus?

src/main/java/dev/dbos/transact/database/SystemDatabase.java
chuck-dbos: In my defense, this diff was caused by an auto rename and those were already there :-)  Will fix.

src/main/java/dev/dbos/transact/database/SystemDatabase.java
chuck-dbos: This is a bit tricky.  `insertWorkflowStatus` throws `SQLException` because it is being passed a connection with an open transaction.

chuck-dbos: In turn, `initWorkflowStatus` needs more help than I want to give it in this PR.

PR #122, opened 2025-10-07
Static `DBOS`

### LINKED ISSUE #98: DBOS.sleep should take a duration

### LINKED ISSUE #99: DBOSContext.dbosInstance should return DBOS instance
Not Optional<DBOS>. It should throw if you access `dbosInstance` and the context instance value is null. Also add a "is there a valid dbos instance method"

### LINKED ISSUE #118: GetWorkflowID and GetStepID
I might have missed them, but I can't find these important methods anywhere in the Java API.

Here's the Go interface, for reference:

https://docs.dbos.dev/golang/reference/methods#dbos-variables

chuck-dbos:
Use a global DBOS, which opens up the registration order flexibility that we have in TS and PY.  (Adding multiple instances would not be that hard, we would pass the DBOS.Instance object around only in initialization and initiation code, everything else would use the context).

Also:
- improve sleep (call sites and use `Duration`)

Fixes #118 
Fixes #99 
Fixes #98

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -44,12 +44,11 @@ plugins {
 
 group = "dev.dbos"
 
-tasks.withType<JavaCompile> {
-    options.release.set(17) // Targets Java 17 bytecode
-}
-
 java {
     withSourcesJar()
+    toolchain {
+        languageVersion.set(JavaLanguageVersion.of(17))
+    }
 }
 
 spotless {
@@ -91,7 +90,7 @@ dependencies {
     testImplementation("io.rest-assured:rest-assured:5.4.0")
     testImplementation("io.rest-assured:json-path:5.4.0")
     testImplementation("io.rest-assured:xml-path:5.4.0")
-    testImplementation(platform("org.junit:junit-bom:5.10.0"))
+    testImplementation(platform("org.junit:junit-bom:5.12.1"))
     testImplementation("org.junit.jupiter:junit-jupiter")
     testImplementation("org.java-websocket:Java-WebSocket:1.5.6")
     testImplementation("org.junit-pioneer:junit-pioneer:2.3.0")
--- a/src/main/java/dev/dbos/transact/conductor/protocol/ListWorkflowsRequest.java
+++ b/src/main/java/dev/dbos/transact/conductor/protocol/ListWorkflowsRequest.java
@@ -137,7 +137,7 @@ public class ListWorkflowsRequest extends BaseMessage {
   public ListWorkflowsInput asInput() {
     var builder =
         new ListWorkflowsInput.Builder()
-            .workflowIDs(body.workflow_uuids)
+            .workflowIds(body.workflow_uuids)
             .workflowName(body.workflow_name)
             .authenticatedUser(body.authenticated_user)
             .startTime(body.start_time != null ? OffsetDateTime.parse(body.start_time) : null)
--- a/src/main/java/dev/dbos/transact/database/WorkflowDAO.java
+++ b/src/main/java/dev/dbos/transact/database/WorkflowDAO.java
@@ -371,7 +371,7 @@ public class WorkflowDAO {
       throw new IllegalStateException("Database is closed!");
     }
 
-    var builder = new ListWorkflowsInput.Builder().workflowIDs(Arrays.asList(workflowId));
+    var builder = new ListWorkflowsInput.Builder().workflowIds(Arrays.asList(workflowId));
     ListWorkflowsInput input = builder.build();
     List<WorkflowStatus> output = listWorkflows(input);
     if (output.size() > 0) {
@@ -399,18 +399,17 @@ public class WorkflowDAO {
     // for mapping to the WorkflowStatus fields by index later in the ResultSet.
     sqlBuilder.append(
         """
-          SELECT workflow_uuid, status, name, recovery_attempts,
-            config_name, class_name, authenticated_user, authenticated_roles,
-            assumed_role, queue_name, executor_id, created_at, updated_at,
-            application_version, application_id,
-            workflow_deadline_epoch_ms, workflow_timeout_ms
+          SELECT workflow_uuid, status, name, config_name, class_name,
+          authenticated_user, assumed_role, authenticated_roles,
+          executor_id, created_at, updated_at, application_version, application_id,
+          recovery_attempts, queue_name, workflow_timeout_ms, workflow_deadline_epoch_ms,
+          started_at_epoch_ms, deduplication_id, priority
         """);
 
     var loadInput = input.loadInput() == null || input.loadInput();
     var loadOutput = input.loadOutput() == null || input.loadOutput();
-
     if (loadInput) {
-      sqlBuilder.append(", inputs, request");
+      sqlBuilder.append(", inputs");
     }
     if (loadOutput) {
       sqlBuilder.append(", output, error");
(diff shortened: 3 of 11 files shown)


src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -326,7 +344,7 @@ public void send(String destinationId, Object message, String topic) {
    * @param timeoutSeconds time in seconds after which the call times out
    * @return the message if there is one or else null
    */
-  public Object recv(String topic, float timeoutSeconds) {
+  public Object recv(String topic, double timeoutSeconds) {
```
devhawk: We should be using Duration for these timeout parameters. Optionally, also provide overloads that take long+TimeUnit params and construct a Duration from Duration.fromNanos

https://github.com/dbos-inc/dbos-transact-java/issues/98

src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -359,7 +377,7 @@ public void setEvent(String key, Object value) {
    * @param timeOut time in seconds to wait for data
    * @return the published value or null
    */
-  public Object getEvent(String workflowId, String key, float timeOut) {
+  public Object getEvent(String workflowId, String key, double timeOut) {
```
devhawk: Duration

src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -377,13 +395,33 @@ public Object getEvent(String workflowId, String key, float timeOut) {
    *
    * @param seconds in seconds
    */
-  public void sleep(float seconds) {
+  public void sleepmsInst(double mseconds) {
```
devhawk: duration, and this is an awful name for a public API

src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -377,13 +395,33 @@ public Object getEvent(String workflowId, String key, float timeOut) {
...
     var executor = dbosExecutor.get();
     if (executor == null) {
       throw new IllegalStateException("cannot sleep before launch");
     }
 
-    executor.sleep(seconds);
+    executor.sleepms(mseconds);
+  }
+
+  public static void sleepms(double mseconds) {
```
devhawk: I don't think we should be taking a single numeric value w/o a timeunit parameter with it

src/main/java/dev/dbos/transact/database/SystemDatabase.java
```diff
@@ -264,10 +262,10 @@ public Object getEvent(
         });
   }
 
-  public double sleep(String workflowId, int functionId, double seconds, boolean skipSleep) {
+  public double sleepms(String workflowId, int functionId, double mseconds, boolean skipSleep) {
```
devhawk: This is an internal API, so I'm not as picky about it. But having the unit of measure in the method name seems much uglier to me than just using Duration all the way thru until it's time to convert to ms to store in the database

src/main/java/dev/dbos/transact/database/StepsDAO.java
```diff
@@ -231,16 +232,16 @@ public List<StepInfo> listWorkflowSteps(String workflowId) throws SQLException {
...
   }
 
-  public double sleep(String workflowUuid, int functionId, double seconds, boolean skipSleep)
+  public double sleep(String workflowUuid, int functionId, Duration duration, boolean skipSleep)
       throws SQLException {
-    return StepsDAO.sleep(dataSource, workflowUuid, functionId, seconds, skipSleep);
+    return StepsDAO.sleep(dataSource, workflowUuid, functionId, duration, skipSleep);
   }
 
   public static double sleep(
```
devhawk: I like what you've done w/ duration parameters, but shouldn't this method be returning a Duration as well?

src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -211,65 +278,95 @@ public Queue build() {
...
-
-    for (Method method : implementation.getClass().getDeclaredMethods()) {
-      Workflow wfAnnotation = method.getAnnotation(Workflow.class);
-      if (wfAnnotation == null) {
-        continue;
+  public static void sleep(Duration duration) {
+    if (!inWorkflow()) {
+      try {
+        Thread.sleep(duration.toMillis());
+      } catch (InterruptedException e) {
```
devhawk, line 367: Pretty sure catch and ignore InterruptedException is a mistake. https://codingtechroom.com/question/understanding-interruptedexception-java

src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -30,143 +32,208 @@
...
-    var executor = dbosExecutor.get();
-    if (executor == null) {
-      throw new IllegalStateException("cannot retrieve workflow before launch");
+      return DBOSInvocationHandler.createProxy(
+          interfaceClass, implementation, instanceName, () -> this.dbosExecutor.get());
     }
 
-    return executor.getWorkflow(className, instanceName, workflowName);
-  }
+    public QueueBuilder Queue(String name) {
```
devhawk, line 112: The readme uses a `DBOS.Queue` method, but I don't see it here

src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -30,143 +32,208 @@
...
 public class DBOS {
+  private DBOS() {}
 
   private static final Logger logger = LoggerFactory.getLogger(DBOS.class);
 
-  private final WorkflowRegistry workflowRegistry = new WorkflowRegistry();
-  private final QueueRegistry queueRegistry = new QueueRegistry();
-  private final List<ScheduledInstance> scheduledWorkflows = new ArrayList<>();
+  public static class Instance {
+    // CB: The registry may deserve a different lifecycle
```
devhawk: I'm not quite there on this approach yet. If we're going to have a global, it feels like we should use it consistently. And it feels weird to require DBOS instance to be created before registration. The proxy only needs to be able to retrieve a dbos executor on method invocation (which is never available at registration time anyway).

src/main/java/dev/dbos/transact/DBOS.java
chuck-dbos: The README was in really bad shape, I started fixing and will finish fixing once we are settled on the rest here.

src/main/java/dev/dbos/transact/DBOS.java
chuck-dbos: We could set thread interrupted again.

src/main/java/dev/dbos/transact/DBOS.java
chuck-dbos: It was already that way... in huddle we talked about improving the registry lifecycle in a separate PR.

chuck-dbos: OK, why not this PR.

src/main/java/dev/dbos/transact/DBOS.java
chuck-dbos: Hopefully left it better than I found it... but we need to keep it on the TODO list.

build.gradle.kts
```diff
@@ -29,7 +29,7 @@ val branchName: String by lazy {
 // Note, this versioning scheme is fine for preview releases
 // but we'll want something more robust once we want to bump
 // the major or minor version number
-val baseVersion = System.getenv("BASE_VERSION") ?: "0.5"
+val baseVersion = System.getenv("BASE_VERSION") ?: "0.6"
```
devhawk: I'm a little antsy about bumping this. 0.5 (w/o preview tag) seems like a good choice for our initial preview. If we advance past 0.5 now, what will we use for our initial release?

PR #131, opened 2025-10-10
Test+Fix getEvent/setEvent

chuck-dbos:
Also do "Id" consistently, and runStep with just name.

### DIFF
--- a/src/main/java/dev/dbos/transact/DBOS.java
+++ b/src/main/java/dev/dbos/transact/DBOS.java
@@ -453,19 +453,19 @@ public class DBOS {
   public static <T, E extends Exception> T runStep(
       ThrowingSupplier<T, E> stepfunc, StepOptions opts) throws E {
 
-    return executor("runStep").runStepI(stepfunc, opts);
+    return executor("runStep").runStepInternal(stepfunc, opts);
   }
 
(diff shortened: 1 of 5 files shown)


src/main/java/dev/dbos/transact/workflow/internal/StepResult.java
```diff
@@ -10,9 +10,9 @@ public class StepResult {
   public StepResult() {}
 
   public StepResult(
-      String workflowId, int functionID, String functionName, String output, String error) {
+      String workflowId, int functionId, String functionName, String output, String error) {
     this.workflowId = workflowId;
-    this.functionId = functionID;
+    this.functionId = functionId;
```
devhawk: nit: should we change these to `stepId`?

src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -456,6 +456,12 @@ public static <T, E extends Exception> T runStep(
     return executor("runStep").runStepI(stepfunc, opts);
   }
 
+  public static <T, E extends Exception> T runStep(ThrowingSupplier<T, E> stepfunc, String name)
+      throws E {
+
+    return executor("runStep").runStepI(stepfunc, new StepOptions(name));
```
devhawk: nit: we probably need a better name than `runStepI`

PR #144, opened 2025-10-16
Fix getResult and getWorkflowStatus to be steps

chuck-dbos:


src/test/java/dev/dbos/transact/workflow/SyncWorkflowTest.java
```diff
@@ -218,18 +218,21 @@ public void multipleChildren() throws Exception {
...
+    assertEquals("child2", steps.get(2).childWorkflowId());
     assertEquals(2, steps.get(2).functionId());
-    assertEquals("childWorkflow3", steps.get(2).functionName());
+    assertEquals("childWorkflow2", steps.get(2).functionName());
+    assertEquals("DBOS.getResult", steps.get(3).functionName());
+
+    assertEquals("child3", steps.get(4).childWorkflowId());
+    assertEquals(4, steps.get(4).functionId());
+    assertEquals("childWorkflow3", steps.get(4).functionName());
+    assertEquals("DBOS.getResult", steps.get(5).functionName());
```
kraftp, line 237: Can we also verify the child workflow ID field is set correctly in these steps? Important for how workflows are visualized

chuck-dbos: You mean getResult is supposed to have child ID in its step record?  (The start workflow ones are checked, as on line 232.)

kraftp: Yeah!

chuck-dbos: Good catch ... it was not doing that.  (But it is now.)

PR #147, opened 2025-10-20
consistent formatting of dbos schema

devhawk:


kraftp (approved):
Does the new syntax properly escape things? Not a huge deal now, but will be necessary for custom schemas

src/main/java/dev/dbos/transact/database/WorkflowDAO.java
```diff
@@ -922,12 +926,13 @@ private static void copyOperationOutputs(
 
     String sql =
         """
-        INSERT INTO dbos.operation_outputs
+        INSERT INTO %1$s.operation_outputs
```
qianl15, line 929: (naive Java question) Do we need `%1$s` here given we only have one string argument? -- I guess yes because it appears multiple times :)

src/main/java/dev/dbos/transact/database/QueuesDAO.java
```diff
@@ -222,8 +220,8 @@ SELECT executor_id, COUNT(*) as task_count
           updatePs.addBatch();
           retIds.add(id);
         }
-        // TODO: should we be checking updateCounts?
-        int[] updateCounts = updatePs.executeBatch();
+
+        updatePs.executeBatch();
```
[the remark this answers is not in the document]
devhawk: we don't check this on TS or PY. Go version has a returning clause so we do check for an error there. However, Unlike all the other languages, Java executes all the updates in a batch instead of one at a time.

src/main/java/dev/dbos/transact/database/WorkflowDAO.java
devhawk: yes, because we use the schema name twice in that query. The full query is:

```sql
        INSERT INTO %1$s.operation_outputs
            (workflow_uuid, function_id, output, error, function_name, child_workflow_id )
        SELECT ? as workflow_uuid, function_id, output, error, function_name, child_workflow_id
            FROM %1$s.operation_outputs
            WHERE workflow_uuid = ? AND function_id < ?
```

PR #148, opened 2025-10-21
Fix Scheduled Workflows

### LINKED ISSUE #117: Scheduled Workflows are Unintuitive
As far as I can tell, to make scheduled workflows work, you have to create a class that only has scheduled workflows (no regular workflows), instantiate it, then call `scheduleWorkflow` on it (but don't at any point create a proxy, or the call will fail with a duplicate registration). This is very unintuitive. Can we make the proxy schedule workflows directly? Or make them work together somehow?

Also, what cron syntax are we using? What's with all the question marks in the examples?

devhawk:
Register scheduled workflows the same way as other workflows are registered. `DBOS.scheduleWorkflow` is removed in favor of 
`DBOS.registerWorkflows`

other changes:
* Use Spring 5.3 cron expression syntax
* Remove `WorkflowFunctionReflect` type. `RegisteredWorkflow` now just holds a `Method` reference which can be used for both annotation reflection and invocation. 
* move SchedulerService and QueueService into execution package
* move Queue into workflow package

fixes #117

### DIFF
--- a/src/main/java/dev/dbos/transact/execution/RegisteredWorkflow.java
+++ b/src/main/java/dev/dbos/transact/execution/RegisteredWorkflow.java
@@ -17,6 +17,7 @@ public record RegisteredWorkflow(
     Objects.requireNonNull(className, "workflow class name must not be null");
     instanceName = Objects.requireNonNullElse(instanceName, "");
     Objects.requireNonNull(target, "workflow target object must not be null");
+    Objects.requireNonNull(workflowMethod, "workflow method must not be null");
   }
 
   public RegisteredWorkflow(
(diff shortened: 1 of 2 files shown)


chuck-dbos:
Love it.
Is use of scheduler state going to be part of this, or a second PR?

devhawk:
> Love it. Is use of scheduler state going to be part of this, or a second PR?

We don't currently support `SchedulerMode.ExactlyOncePerInterval' and @kraftp [doesn't think we need it](https://github.com/dbos-inc/dbos-transact-java/issues/87#issuecomment-3382868811)

devhawk:
> I will interpret this as "second PR"... every time we tell people to use a DB table for their schedule records and put a scheduled workflow in to check it, they need `SchedulerMode.ExactlyOncePerInterval` for correct behavior.

yup, adding scheduler mode will be a 2nd PR post initial preview

PR #149, opened 2025-10-21
Use PMD to detect bogus string compares

chuck-dbos:


build.gradle.kts
```diff
@@ -61,6 +62,20 @@ spotless {
     }
 }
 
+pmd {
```
devhawk, line 65: will the compilation fail if a rule is violated?

chuck-dbos: The `:pmdMain` and `:pmdTest` tasks currently will fail if run when there is a problem.  We would decide how we want to automate runs of those, maybe in GHA?

PR #150, opened 2025-10-21
Uniform use of withers

chuck-dbos:
Provide withers for ListWorkflowsInput, DBOSConfig, Queue, etc.

Configuration behavior also changed... hidden use of env variables, which had been in `build()`, now only occurs if you call `defaultsFromEnv`, which is convenient for tests but not referenced in the README.  We may continue to improve in this area, but it is better than it was.

Also:
  Fix some issues with queue concurrency settings
  Test listWorkflows and listWorkflowSteps when called as steps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,8 +83,8 @@ public class Demo {
         var dbosConfig = DBOSConfig
             .defaults("demo")
             .withDatabaseUrl(System.getenv("DBOS_SYSTEM_JDBC_URL"))
-            .withDbUser(System.getenv("PGUSER"))
-            .withDbPassword(System.getenv("PGPASSWORD"));
+            .withDbUser(Objects.requireNonNullElse(System.getenv("PGUSER"), "postgres"))
+            .withDbPassword(Objects.requireNonNullElse(System.getenv("PGPASSWORD"), "dbos"))
 
         DBOS.configure(dbosConfig);
         


kraftp (commented):
This looks good. Using with instead of a builder improves consistency.

src/main/java/dev/dbos/transact/config/DBOSConfig.java
```diff
@@ -22,7 +22,9 @@ public record DBOSConfig(
     String appVersion,
     String executorId) {
 
-  public DBOSConfig {
+  public DBOSConfig {}
```
devhawk: Why did we remove the custom constructor in favor of a validate method?

src/main/java/dev/dbos/transact/workflow/Queue.java
```diff
@@ -4,23 +4,46 @@
...
     String name,
-    int concurrency,
-    int workerConcurrency,
+    Integer concurrency,
+    Integer workerConcurrency,
     boolean priorityEnabled,
     RateLimit rateLimit) {
 
   public Queue {
     Objects.requireNonNull(name, "Queue name must not be null");
```
devhawk, line 13: Shouldn't we still have the worker concurrency vs concurrency check here? Also, we should also validate that concurrency and worker concurrency are positive if not null

src/main/java/dev/dbos/transact/config/DBOSConfig.java
chuck-dbos: There is no validation until the object is done.  We only know that it is done for sure when it is registered.

src/main/java/dev/dbos/transact/workflow/Queue.java
chuck-dbos: There is no validation until the object is done.  We only know that it is done for sure when it is registered.  That way, if you set them in a weird order, it's OK... as long as the final object is valid.

devhawk: There's never a case for concurrency or workerConcurrency to be zero or negative, so we should check those here.

src/main/java/dev/dbos/transact/config/DBOSConfig.java
devhawk: Yeah, that makes sense. I think we can delete the empty custom constructor then. Also, seems like `DBOS.setConfig` could just check the relevant fields directly rather than delegating that to `DBOSConfig`, but that's a nitpick

src/main/java/dev/dbos/transact/workflow/Queue.java
chuck-dbos: Yes and I can get rid of the checks elsewhere, makes the code there much cleaner.

src/main/java/dev/dbos/transact/config/DBOSConfig.java
```diff
@@ -22,15 +20,28 @@ public record DBOSConfig(
...
-  public DBOSConfig {
-    Objects.requireNonNull(appName, "DBOSConfig.appName must not be null");
-    Objects.requireNonNull(databaseUrl, "DBOSConfig.databaseUrl must not be null");
-    Objects.requireNonNull(dbUser, "DBOSConfig.dbUser must not be null");
-    Objects.requireNonNull(dbPassword, "DBOSConfig.dbPassword must not be null");
-  }
-
   private static final Logger logger = LoggerFactory.getLogger(DBOSConfig.class);
 
+  public static DBOSConfig defaultsFromEnv(String appName) {
```
kraftp: This is nice and concise, but hides too much, I think it's better our sample code be explicit so users know how to change things later (and also so it's clearer what the env vars actually are).

chuck-dbos: So, what could we do?  IMO it was much more hidden before this change, but maybe this isn't explicit enough.

kraftp: Before the change, example code always looked like this:

```java
        DBOSConfig config = new DBOSConfig.Builder()
            .appName("dbos-java-starter")
            .databaseUrl(System.getenv("DBOS_JDBC_URL"))
            .dbUser(System.getenv("PGUSER"))
            .dbPassword(System.getenv("PGPASSWORD"))
            .build();
```

This is very explicit. It will be better/more concise after moving from builders to withers, but replacing the explicit vars with `defaultFromEnv` is not an improvement.

devhawk: Maybe we revert the config change and keep config builder?

chuck-dbos: Fixed the README example to be explicit about reading the env.  Got rid of the automatic env var reads from the builder/withers (except for one method commonly+only used in tests).  This may not be perfect but is better than the prior code on all dimensions discussed.

PR #151, opened 2025-10-21
Remove confusing method; fixes #114

### LINKED ISSUE #114: insertWorkflowStatus needs DB retries
It was not changed because it is the only SystemDatabase method that accepts a `Connection`.  The calling loop needs to be improved, or the boundary of responsibility needs fixed.

chuck-dbos:


src/test/java/dev/dbos/transact/database/SystemDatabaseTest.java
```diff
@@ -1,95 +0,0 @@
...
-    WorkflowStatusInternal wfStatusInternal =
-        new WorkflowStatusInternal(
-            workflowId,
-            WorkflowState.SUCCESS,
-            "OrderProcessingWorkflow",
-            "com.example.workflows.OrderWorkflow",
-            "prod-config",
-            "user123@example.com",
-            "admin",
-            "admin,operator",
```
chuck-dbos, line 65: This is not valid data for the table.  I decided a unit test of INSERT and DELETE statements are no longer of value since we have low-level tests that use the fields properly and would fail reasonably quickly if they broke.

devhawk (approved):

PR #153, opened 2025-10-22
add timeout sum type

### LINKED ISSUE #88: Enable parentTimeoutInheritedByChild
PR #85 changed the way we flow timeout/deadline values. parentTimeoutInheritedByChild needs to be adapted to the new logic

devhawk:
Previously, you would specify a zero timeout to indicate a child workflow had no timeout, even if the parent workflow had a timeout. This was a poor developer experience. 

This PR adds a new `Timeout` sum type that can be one of three values: `Inherit`, `None`, or `Explicit` with `Expicit` including a timeout `Duration`. This allows child worklfows to be started `withTimeout(duration)` or `withNoTimeout()`. Workflow timeout defaults to `Inherit` if not set (which is the same as `None` for top level workflows)

fixes #88

### DIFF
--- a/src/test/java/dev/dbos/transact/database/SystemDatabaseTest.java
+++ b/src/test/java/dev/dbos/transact/database/SystemDatabaseTest.java
@@ -31,12 +31,8 @@ public class SystemDatabaseTest {
   @BeforeAll
   static void onetimeSetup() throws Exception {
     config =
-        new DBOSConfig.Builder()
-            .appName("systemdbtest")
-            .databaseUrl("jdbc:postgresql://localhost:5432/dbos_java_sys")
-            .dbUser("postgres")
-            .maximumPoolSize(3)
-            .build();
+        DBOSConfig.defaultsFromEnv("systemdbtest")
+            .withDatabaseUrl("jdbc:postgresql://localhost:5432/dbos_java_sys");
   }
 
   @BeforeEach


src/main/java/dev/dbos/transact/database/WorkflowDAO.java
```diff
@@ -110,38 +110,31 @@ public WorkflowInitResult initWorkflowStatus(
...
-      } catch (DBOSConflictingWorkflowException | DBOSDeadLetterQueueException e) {
-        // Rollback for custom business exceptions
-        try {
-          connection.rollback();
-        } catch (SQLException rollbackEx) {
-          logger.error("Rollback failed", rollbackEx);
-        }
-        throw e; // Re-throw the custom exception
+      } finally {
+        connection.commit();
```
kraftp, line 136: What happens if a database exception is thrown during the transaction, for example because of a deduplication key violation?

devhawk: I'll add a test to be sure, but `insertWorkflowStatus` throws a DBOSQueueDuplicatedException if it detects a unique contraint violation (aka `23505`). THe tx will commit, but it won't have made a change. This is the same logic as the TS implementation (Python doesn't use a tx in init_workflow)

kraftp: I want to make sure this commit doesn't itself throw an exception--basically, test that the exception thrown from deduplication is the right one.

devhawk: added `SystemDatabaseTest` `testRetries` and `testDedupeId`

PR #158, opened 2025-10-24
Javadocs and small changes

chuck-dbos:
Also:
  WorkflowHandle.getWorkflowId() -> WorkflowHandle.workflowId()
  Withers for ForkOptions
  Remove one OptionalInt as everything similar uses Integer
  Make a private thing private
  Remove a `throws Exception` from something that doesn't throw
  Tested a return value


src/main/java/dev/dbos/transact/context/DBOSContext.java
```diff
@@ -151,11 +150,11 @@ public String getDeduplicationId() {
     return null;
   }
 
-  public OptionalInt getPriority() {
+  public Integer getPriority() {
```
devhawk, line 153: The OptionalInt that can still also be null was a bad choice on my part.

PR #159, opened 2025-10-24
update build version calc

### LINKED ISSUE #74: Improve Build version calculation
we are currently calculating the version number for a builds by using the total commit count as a patch number plus including git short hash and branch name (if not main) in the build metadata. I'm a huge fan of [NerdBank.GitVersioning](https://dotnet.github.io/Nerdbank.GitVersioning/) used by the .NET foundation which we also use in the TS version of DBOS transact, but having a .net runtime build time requirement has been a hassle for DBOS-TS. Want to find a solution that runs on JVM or is a xplat binary (which NBGV unfortunately does not provide)

devhawk:
Maven version order specification: https://maven.apache.org/pom.html#Version_Order_Specification

This PR updates the build version calculation to do the following:

* Calculate major/minor/patch from tag, defaulting to 0.1.0 if no tag is found
* calculate commit count from most recent tag, from start of repo if no tag is foudn
* For main branch, the version is $Major.$Minor + 1.$Patch-m$CommitCount
  * in maven, the m stands for "milestone" which seemed the most reasonable choice out of the pre release choices: alpha/beta/milestone/release candidate 
* For release branch, version is $Major.$Minor.$Patch
  * Additionally, we check to make sure the branch name matches the calculated version and that the branch itself has the tag
  * We might want to enforce these constraints in the publish task, not in the build version calculation
* for any other branch, the version is $Major.$Minor + 1.$Patch-a$CommitCount-gGitHash

fixes #74

### DIFF
--- a/src/main/java/dev/dbos/transact/admin/AdminServer.java
+++ b/src/main/java/dev/dbos/transact/admin/AdminServer.java
@@ -105,7 +105,10 @@ public class AdminServer implements AutoCloseable {
   }
 
   private void healthCheck(HttpExchange exchange) throws IOException {
-    sendJson(exchange, 200, """
+    sendJson(
+        exchange,
+        200,
+        """
         {"status":"healthy"}""");
   }
 
--- a/src/test/java/dev/dbos/transact/admin/AdminServerTest.java
+++ b/src/test/java/dev/dbos/transact/admin/AdminServerTest.java
@@ -230,7 +230,8 @@ class AdminServerTest {
       given()
           .port(port)
           .contentType("application/json")
-          .body("""
+          .body(
+              """
               { "cutoff_epoch_timestamp_ms": 42, "rows_threshold": 37 } """)
           .when()
           .post("/dbos-garbage-collect")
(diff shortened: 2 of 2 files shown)


kraftp (commented):
Let's get this fully documented in DEVELOPING.md, as in Python

kraftp (commented):
The approach looks good. We'll also need infrastructure for generating the tags, like https://github.com/dbos-inc/dbos-transact-py/blob/main/make_release.py in Python.

devhawk:
added docs to DEVELOPING.md + added initial createRelease gradle task

.github/workflows/on_push.yml
```diff
@@ -9,9 +9,9 @@ on:
...
   test:
     uses: ./.github/workflows/test.yml
-  publish:
-    needs: test
-    uses: ./.github/workflows/publish.yml
-    secrets:
-      USERNAME_MAVEN: ${{ secrets.USERNAME_MAVEN }}
-      TOKEN_MAVEN: ${{ secrets.TOKEN_MAVEN }}
\ No newline at end of file
+  # publish:
```
kraftp, line 12: What's the plan here?

devhawk: Update to publish to maven central out of main and release branches, once we get the maven central namespace straightened out

PR #160, opened 2025-10-25
Non file based migrations

### LINKED ISSUE #156: Migrations Should Not Be File-Based
The DBOS migrations are file-based, which adds complexity and potential for breakage. Migrations should work EXACTLY like Python: https://github.com/dbos-inc/dbos-transact-py/blob/main/dbos/_migration.py

Same `dbos_migrations` table, migrations are just strings of SQL code, same text in the same order.

devhawk:
Fixes #156

### DIFF
--- a/src/main/java/dev/dbos/transact/migrations/MigrationManager.java
+++ b/src/main/java/dev/dbos/transact/migrations/MigrationManager.java
@@ -57,18 +57,17 @@ public class MigrationManager {
           }
         }
       } catch (SQLException e) {
-        throw new RuntimeException("Failed to check database", e);
+        logger.warn("SQLException thrown looking for {} database", pair.database(), e);
       }
 
       logger.info("Creating '{}' database", pair.database());
       try (Statement stmt = conn.createStatement()) {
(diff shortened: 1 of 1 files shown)


src/main/java/dev/dbos/transact/migrations/MigrationManager.java
```diff
@@ -85,120 +77,220 @@ public static UrlPair extractDbAndPostgresUrl(String url) {
...
+      var sql =
+          lastApplied == 0
+              ? "INSERT INTO %s.dbos_migrations (version) VALUES (?)"
+              : "UPDATE %s.dbos_migrations SET version = ?";
+      try (var stmt = conn.prepareStatement(sql.formatted(schema))) {
+        stmt.setLong(1, migrationIndex);
+        stmt.executeUpdate();
+      } catch (SQLException e) {
+        throw new RuntimeException("Failed to save migration %d".formatted(migrationIndex), e);
+      }
```
chuck-dbos, line 192: This will probably work most of the time, but I think we can do better.

devhawk: Should we just log the exception and keep going if we can't update the migration version?

PR #194, opened 2025-11-05
retrieveWorkflow in enqueueWorkflow if DBOSWorkflowExecutionConflictException is caught

devhawk:


src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
```diff
@@ -1062,6 +1062,9 @@ public static <T, E extends Exception> WorkflowHandle<T, E> enqueueWorkflow(
           options.timeout(),
           options.deadline());
       return retrieveWorkflow(workflowId, systemDatabase);
+    } catch (DBOSWorkflowExecutionConflictException e) {
+      logger.debug("Workflow execution conflict for workflowId {}", workflowId);
```
kraftp, line 1066: I don't understand how this can happen. That exception should only be thrown from a step, but enqueue should just enqueue, not actually run the workflow or call any steps.

devhawk: It can happen when we enqueue a child workflow

PR #200, opened 2025-11-07
Event receiver infra and scheduler improvements

chuck-dbos:
**Infrastructure for pluggable event receivers**
Add lifecycle listeners and workflow execution entrypoints.
Allow deadline to be specified in start options (consolidates code also)
Test for same.

**Scheduler improvements:**
Make scheduler less special (uses event receivers except for its validator)
Use user-specified queues.
Add mode where missed executions are not ignored, but made up
Fix scheduler's key calculation to be on second boundaries (as it should be)
Check all things in one test, rather than one per run, saving over 10 seconds of test time

### DIFF
--- a/transact/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
+++ b/transact/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
@@ -1111,7 +1111,9 @@ public class DBOSExecutor implements AutoCloseable {
 
     Long timeoutMs = timeout != null ? timeout.toMillis() : null;
     Long deadlineEpochMs =
-        queueName != null ? null : deadline != null ? deadline.toEpochMilli() : null;
+        (queueName != null && timeoutMs != null)
+            ? null
+            : deadline != null ? deadline.toEpochMilli() : null;
 
     final int retries = maxRetries == null ? Constants.DEFAULT_MAX_RECOVERY_ATTEMPTS : maxRetries;
     WorkflowStatusInternal workflowStatusInternal =
--- a/transact/src/test/java/dev/dbos/transact/client/ClientService.java
+++ b/transact/src/test/java/dev/dbos/transact/client/ClientService.java
@@ -4,4 +4,6 @@ public interface ClientService {
   String enqueueTest(int i, String s);
 
   String sendTest(int i);
+
+  void sleep(int ms);
 }
--- a/transact/src/test/java/dev/dbos/transact/client/ClientServiceImpl.java
+++ b/transact/src/test/java/dev/dbos/transact/client/ClientServiceImpl.java
@@ -17,4 +17,13 @@ public class ClientServiceImpl implements ClientService {
     var message = DBOS.recv("test-topic", Duration.ofSeconds(10));
     return String.format("%d-%s", i, message);
   }
+
+  @Workflow
+  public void sleep(int ms) {
+    try {
+      Thread.sleep(ms);
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+    }
+  }
 }
--- a/transact/src/test/java/dev/dbos/transact/client/ClientTest.java
+++ b/transact/src/test/java/dev/dbos/transact/client/ClientTest.java
@@ -10,10 +10,13 @@ import dev.dbos.transact.DBOSClient;
 import dev.dbos.transact.DBOSTestAccess;
 import dev.dbos.transact.config.DBOSConfig;
 import dev.dbos.transact.database.SystemDatabase;
+import dev.dbos.transact.exceptions.DBOSAwaitedWorkflowCancelledException;
 import dev.dbos.transact.utils.DBUtils;
 import dev.dbos.transact.workflow.Queue;
 
 import java.sql.SQLException;
+import java.time.Duration;
+import java.time.Instant;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 
@@ -124,5 +127,40 @@ public class ClientTest {
     assertEquals("SUCCESS", status.status());
 
     assertEquals("42-test.message", handle.getResult());
+  }
+
+  @Test
+  public void clientEnqueueTimeouts() throws Exception {
+    try (var client = new DBOSClient(dbUrl, dbUser, dbPassword)) {
+      var options =
+          new DBOSClient.EnqueueOptions(
+              "dev.dbos.transact.client.ClientServiceImpl", "sleep", "testQueue");
+
+      var handle1 =
+          client.enqueueWorkflow(options.withTimeout(Duration.ofSeconds(1)), new Object[] {10000});
+      assertThrows(
+          DBOSAwaitedWorkflowCancelledException.class,
+          () -> {
+            handle1.getResult();
+          });
+      var stat1 = client.getWorkflowStatus(handle1.workflowId());
+      assertEquals(
+          "CANCELLED",
+          stat1.orElseThrow(() -> new AssertionError("Workflow status not found")).status());
+
+      var handle2 =
+          client.enqueueWorkflow(
+              options.withDeadline(Instant.ofEpochMilli(System.currentTimeMillis() + 1000)),
+              new Object[] {10000});
+      assertThrows(
+          DBOSAwaitedWorkflowCancelledException.class,
+          () -> {
+            handle2.getResult();
+          });
+      var stat2 = client.getWorkflowStatus(handle2.workflowId());
+      assertEquals(
+          "CANCELLED",
+          stat2.orElseThrow(() -> new AssertionError("Workflow status not found")).status());
+    }
   }
 }


transact/src/main/java/dev/dbos/transact/execution/DBOSLifecycleListener.java
```diff
@@ -0,0 +1,13 @@
+package dev.dbos.transact.execution;
+
+/**
+ * For registering callbacks that hear about `DBOS.launch()` and `DBOS.shutdown()`. At this point,
+ * DBOS is ready to run workflows, and no additional registrations are allowed.
+ */
+public interface DBOSLifecycleListener {
+  /** Called from within DBOS.launch, after workflow processing is allowed */
+  void dbosLaunched();
```
devhawk, line 9: do we need "dbos" in the name here? Is the point to differentiate from other framework lifecycle interfaces?

transact/src/main/java/dev/dbos/transact/execution/ExecuteWorkflowOptions.java
```diff
@@ -0,0 +1,56 @@
+package dev.dbos.transact.execution;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Objects;
+
+/** Options for executing a workflow, including identifier, queue, and timeout settings, */
+public record ExecuteWorkflowOptions(
```
devhawk: This seems redundant with StartWorkflowOptions. Why do we need both? Are event listeners allowed to set workflow deadline? Why?

transact/src/main/java/dev/dbos/transact/workflow/Scheduled.java
```diff
@@ -9,4 +9,8 @@
 @Target(ElementType.METHOD)
 public @interface Scheduled {
   String cron();
+
+  String queue() default "";
+
+  boolean ignoreMissed() default true;
```
devhawk, line 15: Oooo, even better than an enum!

transact/src/main/java/dev/dbos/transact/execution/ExecuteWorkflowOptions.java
devhawk: DBOSClient uses ExcecuteWorkflowOptions and DBOSExecutor.enqueueWorkflow because it doesn't have access to the workflow registry. I wouldn't think event receivers are subject to the same constraint

transact/src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -620,6 +651,38 @@ public static List<StepInfo> listWorkflowSteps(String workflowId) {
...
+   * Execute a workflow based on registration and arguments. This is expected to be used by generic
+   * callers, not app code.
+   *
+   * @param regWorkflow Registration of the workflow. @see getRegisteredWorkflows
+   * @param args Workflow function arguments
+   * @param options Execution options, such as ID, queue, and timeout/deadline
+   * @return WorkflowHandle to the executed workflow
+   */
+  public static WorkflowHandle<?, ?> executeWorkflow(
+      RegisteredWorkflow regWorkflow, Object[] args, ExecuteWorkflowOptions options) {
```
devhawk: I'm thinking this should be an overload of startWorkflow. And it should use StartWorkflowOptions. I don't think we should be letting external code set deadlines

transact/src/main/java/dev/dbos/transact/execution/RegisteredWorkflowInstance.java
```diff
@@ -0,0 +1,9 @@
+package dev.dbos.transact.execution;
+
+public record RegisteredWorkflowInstance(
```
devhawk, line 3: I don't understand the use case for this

transact/src/main/java/dev/dbos/transact/execution/DBOSLifecycleListener.java
chuck-dbos: I was on the fence about this, doesn't seem we have a hard and fast rule... happy to make it whichever way.

transact/src/main/java/dev/dbos/transact/execution/ExecuteWorkflowOptions.java
chuck-dbos: It does seem a little redundant.  But I did not invent it, I moved it from DBOSExecutor.

The issue I had was that there are 2 functions - the underlying function and the proxy one.

Calling the proxy one requires a lambda that calls the right method on the proxy.  The scheduler (pre this change) was calling executeWorkflow with the original workflow method, using this structure, and that seemed a lot cleaner, so I opened it up, thinking we could attempt consolidation in a future PR.  I do think the event listeners should bypass the startWorkflow proxy stuff and be allowed to use the method they annotated.

transact/src/main/java/dev/dbos/transact/DBOS.java
chuck-dbos: See above, I left it this way to differentiate it from the proxy version.  Consolidation possibly warranted (attempted consolidation is certainly warranted, just not 100% sure how well it will go).

transact/src/main/java/dev/dbos/transact/execution/RegisteredWorkflowInstance.java
chuck-dbos: This allows objects to be iterated instead of methods.  This would, say, allow a non-workflow to be found (and maybe complained about, like what happens if you put "Scheduled" on a non-workflow... we have the ability to enforce that for Scheduled but only because we know about it before any proxies are created).

TS has a similar thing, but it's probably not strictly necessary.

transact/src/main/java/dev/dbos/transact/DBOS.java
chuck-dbos: I see no reason external code shouldn't be able to set deadlines.  Timeout form and deadline form are quite interchangeable, but why not offer it in an easy way.

Prior to this, we had:
WorkflowOptions
ExecuteWorkflowOptions
StartWorkflowOptions
EnqueueWorkflowOptions

Why not get rid of, at a minimum, ExecuteWorkflowOptions.

transact/src/main/java/dev/dbos/transact/workflow/Scheduled.java
chuck-dbos: IIRC before booleans C programmers all wrote their own enums (or did zero/nonzero).

transact/src/main/java/dev/dbos/transact/StartWorkflowOptions.java
```diff
@@ -4,18 +4,33 @@
...
+ *     priority to assign.
  */
 public record StartWorkflowOptions(
     String workflowId,
     Timeout timeout,
     String queueName,
     String deduplicationId,
-    Integer priority) {
+    Integer priority,
+    Instant deadline) {
```
devhawk, line 33: If we're adding deadline to `StartWorkflowOptions`, we should also add it to `WorkflowOptions`

transact/src/main/java/dev/dbos/transact/StartWorkflowOptions.java
```diff
@@ -27,31 +42,61 @@ public record StartWorkflowOptions(
...
+        this.timeout,
+        this.queueName,
+        this.deduplicationId,
+        this.priority,
+        this.deadline);
   }
 
   /** Produces a new StartWorkflowOptions that overrides timeout value for the started workflow */
   public StartWorkflowOptions withTimeout(Timeout timeout) {
+    if (timeout != null && deadline != null) {
```
devhawk: `this.deadline` so it's clear which is a param and which is not (we do the same in below in this function)

transact/src/main/java/dev/dbos/transact/StartWorkflowOptions.java
```diff
@@ -27,31 +42,61 @@ public record StartWorkflowOptions(
...
+        deadline);
   }
 
   /** Produces a new StartWorkflowOptions that overrides timeout value for the started workflow */
   public StartWorkflowOptions withTimeout(Duration timeout) {
     return withTimeout(Timeout.of(timeout));
   }
 
+  /** Produces a new StartWorkflowOptions that overrides timeout value for the started workflow */
+  public StartWorkflowOptions withTimeoutOrNull(Duration timeout) {
```
devhawk: Isn't this called `withOptionalTimeout` on `WorkflowOptions`? should be consistent

transact/src/main/java/dev/dbos/transact/StartWorkflowOptions.java
```diff
@@ -27,31 +42,61 @@ public record StartWorkflowOptions(
...
+        this.queueName,
+        this.deduplicationId,
+        this.priority,
+        this.deadline);
+  }
+
+  /** Produces a new StartWorkflowOptions that overrides deadline value for the started workflow */
+  public StartWorkflowOptions withDeadline(Instant deadline) {
+    return new StartWorkflowOptions(
+        this.workflowId,
```
devhawk, line 82: We should be checking to ensure timeout and deadline aren't both set here as well. Actually, we should probably just it in the common validator (the code that checks to make sure we don't have a zero or negative timeout - not sure what Java calls that)

transact/src/main/java/dev/dbos/transact/workflow/Timeout.java
```diff
@@ -23,6 +23,11 @@ static Timeout none() {
     return new None();
   }
 
+  static Timeout ofOrNone(Duration d) {
```
devhawk, line 26: Should this just be how `of` works?

chuck-dbos: I would be OK with that - should I change it?

transact/src/main/java/dev/dbos/transact/StartWorkflowOptions.java
chuck-dbos: I tried that, it does fail when the timeout gets promoted to a deadline.  More work could be put into it...

transact/src/main/java/dev/dbos/transact/DBOSClient.java
```diff
@@ -288,14 +288,14 @@ public <T, E extends Exception> WorkflowHandle<T, E> enqueueWorkflow(
...
             Objects.requireNonNullElseGet(options.workflowId(), () -> UUID.randomUUID().toString()),
-            options.timeout(),
-            null,
+            Timeout.ofOrNone(options.timeout()),
             Objects.requireNonNull(
                 options.queueName(), "EnqueueOptions queueName must not be null"),
             options.deduplicationId,
-            options.priority),
+            options.priority,
+            null),
```
devhawk: should we add deadline to Client's EnqueueOptions?

transact/src/main/java/dev/dbos/transact/StartWorkflowOptions.java
chuck-dbos: Coalescing with the comment below, maybe there should be just one.

transact/src/main/java/dev/dbos/transact/StartWorkflowOptions.java
devhawk: lets open an issue to track

PR #225, opened 2025-11-21
update logging in executor and sleep

devhawk:
To run down a cloud failure issue

### DIFF
--- a/transact/src/main/java/dev/dbos/transact/database/StepsDAO.java
+++ b/transact/src/main/java/dev/dbos/transact/database/StepsDAO.java
@@ -283,14 +283,15 @@ public class StepsDAO {
 
     if (recordedOutput != null) {
       logger.debug(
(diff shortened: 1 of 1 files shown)


transact/src/main/java/dev/dbos/transact/database/StepsDAO.java
```diff
@@ -282,14 +282,15 @@ public static Duration sleep(
     double endTime;
 
     if (recordedOutput != null) {
-      logger.debug("Replaying sleep, id: {}, millis: {}", functionId, duration.toMillis());
+      logger.debug(
+          "Replaying sleep, workflow {}, id: {}, millis: {}", workflowUuid, functionId, duration);
```
chuck-dbos: Is it still called labeled `millis`?

devhawk: fixed

PR #228, opened 2025-11-30
handle non serializable exceptions

erkieh:
I had failures where one of my used libraries threw an exception which was not serializable. This change should make handling of these better.


erkieh:
It was a `org.zalando.problem.ThrowableProblem`
which had a non-serializable io.micronaut.problem.HttpStatusType

My biggest problem with that was that once such an exception occurred, the result was not recorded in the workflow status table and the getResult method call was left hanging. This can be reproduced if the AsyncWorkflowTest in the PR is run without my proposed fix in JSONUtil.

Issue #227, opened 2025-11-23
Failing to start workflow when a large amount is done


erkieh:
I had situation where there was a failure when a large amount of worflows was started. I managed to reproduce it in one of your test with some modifications as seen at:

https://github.com/erkieh/dbos-transact-java/commit/6f7655133619dcf113a4c27dbba105d1e60c747c

The error was 
Fail testQueuedMultipleWorkflows with "[7.726s][warning][os,thread] Failed to start thread "Unknown thread" - pthread_create failed (EAGAIN) for attributes: stacksize: 2048k, guardsize: 16k, detached."

It seems that the poole size is not limited and one can run out of resources in some cases.

I also received some OutOfMemory errors before increasing my limits XMX limits.

Please confirm if i have not made any false assumptions about the usage and the ticket is valid.

chuck-dbos:
Will confer w/ @devhawk about the PR.  There may be something to be gained from giving more control to how the executor is set up, whether we want to make it precisely a `java.util.cojava.util.concurrent.ExecutorService`, something more abstract, or just a configuration.  Or if simply improving the one that is there will remove the benefit of configurability / pluggability.

When you provided the executor, what did you change about it?

2048k is an excessive stack size for Java, which passes everything by references.

erkieh:
I used the default 

Executors.newVirtualThreadPerTaskExecutor()

this was enough.

devhawk:
FYI, #234 should address this w/o changing the executor. We are going to investigate using newVirtualThreadPerTaskExecutor when running on Java 21 and later, but that will be a separate PR.

PR #229, opened 2025-11-30
Make Workflow Executorservice Pluggable

erkieh:
This makes it possible to work around the issue in
https://github.com/dbos-inc/dbos-transact-java/issues/227

Also allows the endusers to provide their own Executorservice. Which in my case was a Micrometer wrapped VirtualThreadPerTaskExecutor

### DIFF
--- a/transact/src/main/java/dev/dbos/transact/config/DBOSConfig.java
+++ b/transact/src/main/java/dev/dbos/transact/config/DBOSConfig.java
@@ -2,6 +2,8 @@ package dev.dbos.transact.config;
 
 import dev.dbos.transact.Constants;
 
+import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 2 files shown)


erkieh:
> I need to give this more thought. I'm not a fan of pluggable executors - I'd much rather have a thread executor story for JDK 17, virtual thread executor story for JDK 21+. Additionally, we have to consider how DBOSContext flows. Currently, it uses thread local storage. I'm guessing that needs to change for virtual threads.

Do you want me to close this PR? If the problem gets resolved I am happy either way.

devhawk:
Closing. #234 resolved the issue from #227. Opened #258 to track using VirtualThreadPerTaskExecutor  on Java 21+

PR #230, opened 2025-11-30
Make JSON serde pluggable

erkieh:
I needed to register custom Jackson modules (Java8, Kotlin) and Mixins (For some other external library Classes which were not serializable by the default Jackson implementation). 

Also I might need to use the Micronaut Serde which is compatible with Jackson annotations and has the benefit of being GraalVM native compatible.

So I created this PR for your consideration.

### DIFF
--- a/transact/src/main/java/dev/dbos/transact/json/JSONSerde.java
+++ b/transact/src/main/java/dev/dbos/transact/json/JSONSerde.java
@@ -1,0 +1,11 @@
+package dev.dbos.transact.json;
+
+public interface JSONSerde {
+  String serializeArray(Object[] args);
+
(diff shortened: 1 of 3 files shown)


erkieh:
> I like the idea here, but I think we need to plumb this thru `DBOSConfig` rather than a setter on JSONUtil. Plus, we need tests

it's obvious now that you say it :)

Would you guys like to implement this feature yourselves or I should work on this PR? It does not make difference to me. I can't give promises when I would be able to finish it.

devhawk:
Do we want this to be pluggable or do we want to switch to a more robust library like Micronaut Serde?

PR #232, opened 2025-12-01
getEvent + recv poll occasionally

### LINKED ISSUE #203: recv / getEvent wait until timeout if notifier is disrupted
The systemdb implementation of recv / getEvent never polls, so if the listener is disrupted when the notification arrives, the recv/getEvent call hangs until its timeout is reached.

chuck-dbos:
Fixes #203

### DIFF
--- a/transact/src/main/java/dev/dbos/transact/database/NotificationsDAO.java
+++ b/transact/src/main/java/dev/dbos/transact/database/NotificationsDAO.java
@@ -33,7 +33,7 @@ public class NotificationsDAO {
     this.notificationService = nService;
   }
 
-  public void speedUpPollingForTest() {
+  void speedUpPollingForTest() {
     dbPollingIntervalEventMs = 100;
   }
 
--- a/transact/src/main/java/dev/dbos/transact/database/SystemDatabase.java
+++ b/transact/src/main/java/dev/dbos/transact/database/SystemDatabase.java
@@ -75,7 +75,7 @@ public class SystemDatabase implements AutoCloseable {
     notificationService.stop();
   }
 
-  public void speedUpPollingForTest() {
+  void speedUpPollingForTest() {
     workflowDAO.speedUpPollingForTest();
     notificationsDAO.speedUpPollingForTest();
   }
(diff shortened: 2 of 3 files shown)


transact/src/main/java/dev/dbos/transact/database/NotificationsDAO.java
```diff
@@ -25,13 +25,18 @@ public class NotificationsDAO {
...
   private NotificationService notificationService;
+  private long dbPollingIntervalEventMs = 10000;
 
   NotificationsDAO(HikariDataSource ds, NotificationService nService, String schema) {
     this.dataSource = ds;
     this.schema = Objects.requireNonNull(schema);
     this.notificationService = nService;
   }
 
+  public void speedUpPollingForTest() {
```
devhawk: Does this have to be public?

transact/src/main/java/dev/dbos/transact/database/SystemDatabase.java
```diff
@@ -75,18 +75,9 @@ public void stop() {
...
-   * @param workflowId The workflow ID
-   * @return Optional containing the raw output string if workflow completed successfully, empty
-   *     otherwise
-   */
-  public Optional<String> getWorkflowResult(String workflowId) {
-    return DbRetry.call(
-        () -> {
-          return workflowDAO.getWorkflowResult(workflowId);
-        });
+  public void speedUpPollingForTest() {
```
devhawk: actually, it's this one that should not be public

chuck-dbos: Not sure about the merits of that ... if the methods on an object are public (so they can be called from other packages) but you can't get a reference to the object by public means?

devhawk: yeah, good point.

chuck-dbos: I will change it, we can make it public later if we want to use it in tests for other packages.

PR #236, opened 2025-12-05
Single Execution

### LINKED ISSUE #226: Starting Workflow with the same workFlowId causes Workflow to run multiple times concurrently
When starting a workflow with an external provided workflowId such as:

```java
StartWorkflowOptions options = new StartWorkflowOptions().withWorkflowId(String.valueOf(request.num()));
DBOS.startWorkflow(() -> personService.insert(request.num(), request.path()), options);
```

some steps are executed multiple times concurrently when starting the workflow multiple times.

For example, with the following workflow definition:

```java
@Workflow(name = "insert")
public void insert(int num, String path) {
    System.out.println("Starting Insert for num %d".formatted(num));
    DBOS.runStep(() -> createTable(num), "Create Table");
    DBOS.sleep(Duration.ofSeconds(30));
    DBOS.runStep(() -> insertData(num, path), "Insert Data");
}
```

and starting the workflow twice, you get the following behaviour:

```
Starting Insert for num 14
Running Step: Create Table
2025-11-23T10:08:22.781+01:00  INFO 24152 --- [pool-2-thread-2] com.zaxxer.hikari.HikariDataSource       : HikariPool-4 - Starting...
2025-11-23T10:08:22.790+01:00  INFO 24152 --- [pool-2-thread-2] com.zaxxer.hikari.pool.HikariPool        : HikariPool-4 - Added connection org.postgresql.jdbc.PgConnection@588080ea
2025-11-23T10:08:22.791+01:00  INFO 24152 --- [pool-2-thread-2] com.zaxxer.hikari.HikariDataSource       : HikariPool-4 - Start completed.
Starting Insert for num 14
Running Step: Insert Data
Read CSV
Running Step: Insert Data
Read CSV
Insert Data
Insert Data
```

Based on the following sentence from the documentation, I would have expected the second start to do nothing:

> An assigned workflow ID acts as an idempotency key: if a workflow is called multiple times with the same ID, it executes only once.

chuck-dbos:
Use database to limit workflows to one executor globally.
Use memory map to limit to one execution locally.
Close commit holes in WF status and step outputs.
Test infrastructure and tests.
Fixes #226

### DIFF
--- a/transact/src/main/java/dev/dbos/transact/DBOSClient.java
+++ b/transact/src/main/java/dev/dbos/transact/DBOSClient.java
@@ -381,7 +381,7 @@ public class DBOSClient implements AutoCloseable {
     var status =
         new WorkflowStatusInternal(workflowId, WorkflowState.SUCCESS)
             .withName("temp_workflow-send-client");
-    systemDatabase.initWorkflowStatus(status, null, false, false, UUID.randomUUID().toString());
+    systemDatabase.initWorkflowStatus(status, null, false, false);
     systemDatabase.send(status.workflowId(), 0, destinationId, message, topic);
   }
 
--- a/transact/src/main/java/dev/dbos/transact/database/SystemDatabase.java
+++ b/transact/src/main/java/dev/dbos/transact/database/SystemDatabase.java
@@ -85,6 +85,10 @@ public class SystemDatabase implements AutoCloseable {
    *
    * @param initStatus The initial workflow status details.
    * @param maxRetries Optional maximum number of retries.
+   * @param isRecoveryRequest True if this is a recovery request, indicating that this node is told
+   *     it owns the workflow even if the ID already exists
+   * @param isDequeuedRequest True if this is a dequeue request, indicating that this node is told
+   *     it owns the workflow (provided it is in the enqueued state)
    * @return An object containing the current status and optionally the deadline epoch milliseconds.
    * @throws DBOSConflictingWorkflowException If a conflicting workflow already exists.
    * @throws DBOSMaxRecoveryAttemptsExceededException If the workflow exceeds max retries.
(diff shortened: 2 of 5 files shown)


transact/src/main/java/dev/dbos/transact/database/WorkflowDAO.java
```diff
@@ -100,14 +122,30 @@ public WorkflowInitResult initWorkflowStatus(
...
+          connection.commit();
+        } else {
+          connection.rollback();
+        }
+        DebugTriggers.debugTriggerPoint(DebugTriggers.DEBUG_TRIGGER_INITWF_COMMIT);
       }
     } // end try with resources connection closed
   }
 
+  public static record InsertWorkflowResult(
```
devhawk: nit: I don't think this needs to be public or static

transact/src/main/java/dev/dbos/transact/migrations/MigrationManager.java
```diff
@@ -394,4 +394,9 @@ FOREIGN KEY (workflow_uuid) REFERENCES %1$s.workflow_status(workflow_uuid)
       );
       ALTER TABLE %1$s.streams ADD COLUMN function_id INTEGER NOT NULL DEFAULT 0;
       """;
+
+  static final String migration7 =
+      """
+      ALTER TABLE %1$s."workflow_status" ADD COLUMN "owner_xid" VARCHAR(40) DEFAULT NULL
```
devhawk, line 400: Q: What does xid stand for here? Is it executor ID?

transact/src/main/java/dev/dbos/transact/database/WorkflowDAO.java
```diff
@@ -116,7 +154,11 @@ public WorkflowInitResult initWorkflowStatus(
    * @throws SQLException
    */
   public InsertWorkflowResult insertWorkflowStatus(
```
devhawk: nit: does insertWorkflowStatus need to be public?

transact/src/main/java/dev/dbos/transact/database/WorkflowDAO.java
chuck-dbos: Not a single thing in any DAO needs to be public, I was just blending in.  `static` should be your default unless there is a good reason to have access to an instance of the outer class... which this one does not need.

transact/src/main/java/dev/dbos/transact/migrations/MigrationManager.java
chuck-dbos: Attempt.

chuck-dbos: Discussed w/ @devhawk and improved doc at systemdatabase level.

PR #245, opened 2025-12-09
partition queues

### LINKED ISSUE #206: Partitioned Queues
We should also implement these in Java: https://docs.dbos.dev/python/tutorials/queue-tutorial#partitioning-queues

devhawk:
fixes #206

### DIFF
--- a/transact/src/test/java/dev/dbos/transact/client/ClientTest.java
+++ b/transact/src/test/java/dev/dbos/transact/client/ClientTest.java
@@ -90,11 +90,6 @@ public class ClientTest {
     assertThrows(IllegalArgumentException.class, () -> options.withDeduplicationId(""));
     assertThrows(IllegalArgumentException.class, () -> options.withQueuePartitionKey(""));
 
-    // dedupe ID and partition key must not both be set
-    assertThrows(
-        IllegalArgumentException.class,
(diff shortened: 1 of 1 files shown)


transact/src/main/java/dev/dbos/transact/StartWorkflowOptions.java
```diff
@@ -23,14 +23,17 @@
...
  */
 public record StartWorkflowOptions(
     String workflowId,
     Timeout timeout,
     Instant deadline,
     String queueName,
     String deduplicationId,
-    Integer priority) {
+    Integer priority,
+    String queuePartitionKey) {
```
chuck-dbos, line 36: I ask if this is part of the public API?  (A bit early to worry about it, and not the way we recommend doing things.)

devhawk: I'm not sure what "this" you're referring to, but `StartWorkflowOptions` and `queuePartitionKey` are both parts of the public API. Adding a field to a record like this is a breaking change for anyone calling the all fields ctor - but I would expect folks to primarily call the default or workflowId ctors and use the with* methods rather than calling the all fields ctor.

kraftp (approved):
This looks correct to me

PR #249, opened 2025-12-15
patch support

devhawk:


transact/src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -752,4 +752,34 @@ public static Optional<ExternalState> getExternalState(
...
   }
+
+  /**
+   * Apply a patch within a workflow execution to handle code changes and updates. This method
+   * allows workflows to conditionally execute different code paths based on whether a specific
+   * patch has been applied. Patching must be enabled in DBOS configuration and this method must be
+   * called from within a workflow context.
+   *
+   * @param patchName the name of the patch to apply
+   * @return true if the patch was successfully applied, false if it was already applied previously
```
chuck-dbos: That is not what it does.

devhawk: fixed

transact/src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -752,4 +752,35 @@ public static Optional<ExternalState> getExternalState(
...
   }
+
+  /**
+   * Marks a breaking change within a workflow. Returns true for workflows started after the
+   * breaking change, false for worklows started before the breaking change. Patching must be
+   * enabled in DBOS configuration and this method must be called from within a workflow context.
+   *
+   * @param patchName the name of the patch to apply
+   * @return true for workflows started after the breaking change, false for workflows started
+   *     before the breaking change
```
chuck-dbos, line 766: I don't like this either.  "started before" is not technically accurate, as it is "hit this spot before". Returns true if the code should take the patch branch, false if it is a reexecution that should not take the patch branch.

transact/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
```diff
@@ -738,6 +752,46 @@ public <T, E extends Exception> T getResult(String workflowId) throws E {
...
+    var patched = systemDatabase.patch(workflowId, functionId, patchName);
+    if (patched) {
+      ctx.getAndIncrementFunctionId();
+    }
+    return patched;
+  }
+
+  public boolean deprecatePatch(String patchName) {
+    if (!config.enablePatching()) {
+      throw new RuntimeException("Patching must be enabled in DBOS Config");
```
chuck-dbos: Suggest `IllegalStateException` as more specific and still appropriate, also see above, and test.

PR #257, opened 2026-01-06
listenerQueues

### LINKED ISSUE #244: Implement Explicit Queue Listening
Analogous to https://github.com/dbos-inc/dbos-transact-py/pull/537

devhawk:
fixes #244

### DIFF
--- a/transact/src/test/java/dev/dbos/transact/queue/QueuesTest.java
+++ b/transact/src/test/java/dev/dbos/transact/queue/QueuesTest.java
@@ -22,7 +22,6 @@ import dev.dbos.transact.workflow.internal.WorkflowStatusInternal;
 import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.SQLException;
-import java.time.Duration;
 import java.time.OffsetDateTime;
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


transact/src/test/java/dev/dbos/transact/queue/QueuesTest.java
```diff
@@ -660,4 +659,28 @@ public void testQueueConcurrencyUnderRecovery() throws Exception {
...
+
+    var h1 =
+        DBOS.startWorkflow(
+            () -> serviceQ.simpleQWorkflow("one"), new StartWorkflowOptions(queueOne));
+    var h2 =
+        DBOS.startWorkflow(
+            () -> serviceQ.simpleQWorkflow("two"), new StartWorkflowOptions(queueTwo));
+
+    assertEquals("oneone", h1.getResult());
+    assertEquals("ENQUEUED", h2.getStatus().status());
```
chuck-dbos, line 685: It seems to me that this test relies on the passage of time, the longer it goes the more faith we have that the workflow would "never" get dequeued.  So ... should we enqueue 2 earlier than 1 then?

devhawk: sure, and I'll add a delay

PR #259, opened 2026-01-06
use newVirtualThreadPerTaskExecutor on Java 21

### LINKED ISSUE #258: use newVirtualThreadPerTaskExecutor on Java 21
originally raised as part of #227. Fix for that issue didn't require `newVirtualThreadPerTaskExecutor` but should still consider this for Java 21+

devhawk:
Updates the executor used for executing workflows:
* use `newVirtualThreadPerTaskExecutor` where available
* use `newFixedThreadPool` if virtual thread per task executor is unavailable. 
  * fixed thread pool thread count limited to cpu count * 50

Fixes #258

~Also changes all tests (except conductor tests) to use `@Test` instead of `@RetryingTest(3)`~

### DIFF
--- a/transact/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
+++ b/transact/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
@@ -240,7 +240,7 @@ public class DBOSExecutor implements AutoCloseable {
       queueService.stop();
       queueService = null;
       systemDatabase.stop();
-      if (config.dataSource() != null) {
+      if (config.dataSource() == null) {
         systemDatabase.close();
       }
       systemDatabase = null;


transact/src/test/java/dev/dbos/transact/execution/ScaleTest.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  @Test
+  public void scaleTest() throws Exception {
+    var service = DBOS.registerWorkflows(ScaleService.class, new ScaleServiceImpl());
+    DBOS.launch();
+
+    var usingThreadPoolExecutor = DBOSTestAccess.getDbosExecutor().usingThreadPoolExecutor();
+    final int count =
+        Runtime.getRuntime().availableProcessors() * (usingThreadPoolExecutor ? 50 : 500) * 2;
```
chuck-dbos: This adds 2 minutes to the total runtime of the test suite (on my machine).  (It can't add any less than one minute.)  This is a valuable test, but it is for something that is not going to change (risk breakage) very often.  Can we either speed it up (have them not wait the full 60 seconds once they are all launched) or run the test in a different suite?

chuck-dbos: Another thing I'd like to see is that `usingThreadPoolExecutor` is returning the correct value.  Otherwise, we have not verified that the new code path is ever used.

devhawk: added tests to verify thread pool type on a per java version basis. Investigating disabling scaleTest by default and only running in CI

PR #265, opened 2026-01-13
Supplier<ExecutorService>

devhawk:
Modifies the `DBOSExecutor.executorService` to use `newVirtualThreadPerTaskExecutor` if available, falling back to use `newFixedThreadPool` instead of `newCachedThreadPool`.

Other changes:
* Updated `ClientTest.clientEnqueueTimeouts` and `QueuesTest.testQueueConcurrencyUnderRecovery` to be more reliable 
* updated a bunch of tests that were not closing down data sources correctly
* Updated build.gradle.kts to use environment JDK for tests instead of downloading JDK 17

### DIFF
--- a/transact/src/test/java/dev/dbos/transact/queue/ConcurrencyTestServiceImpl.java
+++ b/transact/src/test/java/dev/dbos/transact/queue/ConcurrencyTestServiceImpl.java
@@ -24,9 +24,9 @@ public class ConcurrencyTestServiceImpl implements ConcurrencyTestService {
 
   @Workflow(name = "blockedWorkflow")
   public int blockedWorkflow(int i) throws InterruptedException {
+    counter.incrementAndGet();
     logger.info("release {} semaphore", i);
     wfSemaphore.release();
-    counter.incrementAndGet();
     latch.await();
     return i;
   }


transact/src/test/java/dev/dbos/transact/client/ClientTest.java
```diff
@@ -189,10 +189,8 @@ public void clientEnqueueTimeouts() throws Exception {
           "CANCELLED",
           stat1.orElseThrow(() -> new AssertionError("Workflow status not found")).status());
 
-      var handle2 =
-          client.enqueueWorkflow(
-              options.withDeadline(Instant.ofEpochMilli(System.currentTimeMillis() + 1000)),
-              new Object[] {10000});
+      var deadline = Instant.ofEpochMilli(System.currentTimeMillis());
+      var handle2 = client.enqueueWorkflow(options.withDeadline(deadline), new Object[] {10000});
```
chuck-dbos: This was an intentional change to save 1 second on the test (but possibly cause the cancel point to move)?

devhawk: the change was to ensure the deadline would hit, but looking at it again I'm not sure I like this change.

devhawk: I reverted this change

PR #267, opened 2026-01-14
DbosKotlinInternal class

devhawk:
Initial experiment for including a wrapper around the various `DBOS.*` static methods to make them kotlin friendly. In particular:

* wrap all DBOS static methods as top level functions
* For methods that take a lambda like `runStep` and `startWorkflow`, make the lambda the last parameter to take advantage of [trailing lambda syntax](https://kotlinlang.org/docs/lambdas.html#passing-trailing-lambdas)
* for methods that take a generic Exception parameter, hard code to `Exception` as Kotlin doesn't have checked exceptions
* use default params where possible (such as `registerWorkflows` `instanceName` and `send` `idempotencyKey`)


chuck-dbos:
This looks harmless, maybe we merge it and see if it works easily in the demo then?

PR #280, opened 2026-02-04
Portable JSON serializer (Java)

chuck-dbos:
Allow Java to interoperate with TS/Py via simple JSON serialization to the workflow_status and other tables

### DIFF
--- a/transact/src/main/java/dev/dbos/transact/migrations/MigrationManager.java
+++ b/transact/src/main/java/dev/dbos/transact/migrations/MigrationManager.java
@@ -447,11 +447,11 @@ public class MigrationManager {
 
   static final String migration11 =
       """
-      ALTER TABLE %1$s."workflow_status" ADD COLUMN "serialization" TEXT DEFAULT NULL;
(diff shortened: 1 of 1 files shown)


transact/src/main/java/dev/dbos/transact/migrations/MigrationManager.java
```diff
@@ -407,4 +408,14 @@ FOREIGN KEY (workflow_uuid) REFERENCES %1$s.workflow_status(workflow_uuid)
       ALTER TABLE %1$s."workflow_status" ADD COLUMN "parent_workflow_id" TEXT DEFAULT NULL;
       CREATE INDEX "idx_workflow_status_parent_workflow_id" ON %1$s."workflow_status" ("parent_workflow_id");
       """;
+
+  static final String migration9 =
```
devhawk, line 420: this does not match Python's migration nine

merryoncherry: It matched when it was written.  Just be aware that you do not want to reorder migrations once they are merged to main because of the limited way they are tracked.  They need to reach the same end state.

devhawk (approved):

PR #281, opened 2026-02-05
WorkflowClassName annotation

chuck-dbos:


devhawk (approved):

devhawk (changes requested):
I approved, but then thought of something: what do we do if multiple classes declare themselves to have the same name?

devhawk (approved):
the workflow registry already requires each workflow to have a unique class+instance+workflow name. If we alias two classes to have the same name, that's OK so long as none of the workflow method names collide

PR #287, opened 2026-02-23
Fixes custom schema processing

devhawk:
The `sanitizeSchema` method originally added changed `"` to `""` and added double quotes around the entire schema so that it would support schema names with single and double quotes in them such as `C$+0m'`. However, this approach is not compatible with Migration #10, where we need to use the schema name as a column value.

### DIFF
--- a/transact/src/main/java/dev/dbos/transact/database/SystemDatabase.java
+++ b/transact/src/main/java/dev/dbos/transact/database/SystemDatabase.java
@@ -51,11 +51,12 @@ public class SystemDatabase implements AutoCloseable {
   private final NotificationService notificationService;
 
   private SystemDatabase(DataSource dataSource, String schema, boolean created) {
-    this.schema = sanitizeSchema(schema);
+    schema = sanitizeSchema(schema);
     if (schema.contains("'") || schema.contains("\"")) {
       throw new IllegalArgumentException("Schema name must not contain single or double quotes");
     }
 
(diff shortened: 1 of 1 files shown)


transact/src/main/java/dev/dbos/transact/database/SystemDatabase.java
```diff
@@ -37,11 +37,9 @@ public class SystemDatabase implements AutoCloseable {
...
 
   public static String sanitizeSchema(String schema) {
-    schema =
-        Objects.requireNonNullElse(schema, Constants.DB_SCHEMA)
-            .replace("\0", "")
-            .replace("\"", "\"\"");
-    return "\"%s\"".formatted(schema);
+    return Objects.requireNonNullElse(schema, Constants.DB_SCHEMA)
+        .replace("\0", "")
+        .replace("\"", "\"\"");
```
devhawk: I'm thinking here we should just remove single and double quotes, like we remove the null char. This will allow the schema to be used as a schema in queries as well as a value in Migration #10

kraftp: I'd prefer it just fail outright with a clear error message

devhawk: I added an explicit check for schemas with a single or double quote in them

PR #291, opened 2026-02-28
DBOS Instance support

devhawk:
ensures DBOS.Instance is usable and that all DBOS statics use the global DBOS instance.

* DBOS.Instance is publicly constructable. 
  * DBOS.Instance.setConfig removed - config is passed to the ctor
* added a DBOS.Instance parameter to DBOSLifecycleListener.launch
* updated DBOS internals to use DBOS.Instance instead of DBOS statics
  * changed WorkflowHandleFuture and WorkflowHandleDBPoll to use an executor instance
  * Added ClientWorkflowHandle that uses a sys db instance
* DBOS.globalInstance is now an AtomicReference, eliminating the need for `syncronized` methods to set it
* DBOS.configure sets the global instance if it is not null
* DBOS.reinitialize unconditionally sets the global instance. This method is now package private as it is primarily intended for our test usage
* DBOS static methods all forward to the global instance. 
  * The global instance is *NOT* created on demand. If it is not set via DBOS.configure (or reinitialize) the static methods fail
* Migrations are run on launch instead of when calling DBOS.configure
* hard coded internal queue support in queue registry to eliminate need to register internal queue
* minor cleanups (registerQueue no longer returns the queue, registerWorkflowMethod no longer returns the method's name)
* added JavaDocs to all the DBOS.Instance public methods

Note: currently, there's no DBOS.Instance mechanism for something like using DBOS.startWorkflow to start a child workflow. Opened #296 to track fixing this.

### DIFF
--- a/transact/src/main/java/dev/dbos/transact/DBOS.java
+++ b/transact/src/main/java/dev/dbos/transact/DBOS.java
@@ -945,8 +945,8 @@ public class DBOS {
   /**
    * Get the serialization format of the current workflow context.
    *
-   * @return the serialization format name (e.g., "portable_json", "java_jackson"), or null if not
-   *     in a workflow context or using default serialization
+   * @return the SerializationStrategy (e.g., "portable_json", "java_jackson"), or null if not in a
+   *     workflow context or using default serialization
    */
   public static @Nullable SerializationStrategy serializationStrategy() {
     return DBOSContext.serializationStrategy();
--- a/transact/src/main/java/dev/dbos/transact/internal/DBOSInvocationHandler.java
+++ b/transact/src/main/java/dev/dbos/transact/internal/DBOSInvocationHandler.java
@@ -46,6 +46,7 @@ public class DBOSInvocationHandler implements InvocationHandler {
             new DBOSInvocationHandler(implementation, instanceName, executor));
   }
 
+  @Override
   public Object invoke(Object proxy, Method method, Object[] args) throws Exception {
 
     var implMethod = target.getClass().getMethod(method.getName(), method.getParameterTypes());
@@ -98,15 +99,14 @@ public class DBOSInvocationHandler implements InvocationHandler {
             : target.getClass().getName();
     var workflowName = workflow.name().isEmpty() ? method.getName() : workflow.name();
     var executor = executorSupplier.get();
+    if (executor == null) {
+      throw new IllegalStateException("executorSupplier returned null");
+    }
 
     if (hook != null) {
       var invocation = new Invocation(executor, className, instanceName, workflowName, args);
       hook.invoke(invocation);
       return defaultReturn(method);
-    }
-
-    if (executor == null) {
-      throw new IllegalStateException("executorSupplier returned null");
     }
 
     var handle = executor.invokeWorkflow(className, instanceName, workflowName, args);


transact/src/test/java/dev/dbos/transact/invocation/InstanceTest.java
```diff
@@ -0,0 +1,511 @@
...
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class HawkServiceInstanceImpl implements HawkService {
+  private final DBOS.Instance dbos;
+  private HawkService proxy;
+
+  public HawkServiceInstanceImpl(DBOS.Instance dbos) {
+    this.dbos = dbos;
```
kraftp, line 37: This is never used--maybe one of the test workflows should use a method like DBOS.sleep or DBOS.setEvent?

transact/src/test/java/dev/dbos/transact/invocation/InstanceTest.java
```diff
@@ -0,0 +1,511 @@
...
+
+  @Workflow
+  @Override
+  public String illegalWorkflow() {
+    return proxy.illegalStep();
+  }
+}
+
+@org.junit.jupiter.api.Timeout(value = 2, unit = java.util.concurrent.TimeUnit.MINUTES)
+public class InstanceTest {
```
kraftp, line 115: I may have missed it, but there should be a test that we can create multiple instances and they should all work, unless that's not possible? Is the instance still a singleton?

kraftp (commented):
This makes sense. Having to pass the instance around is annoying, but unavoidable, and like you said there are ways to improve it in the future. And this will still be optional, you can still use the static if you want?  

What I don't understand is if the instance remains a singleton, and what happens if you try to create multiple.

Also we should probably do a quick test mocking the instance, to verify it works as we expect as that's one of the main use cases for this feature.

transact/src/test/java/dev/dbos/transact/invocation/InstanceTest.java
devhawk: the instance test uses it

transact/src/test/java/dev/dbos/transact/invocation/InstanceTest.java
devhawk: yes, we need tests for this scenario.

devhawk:
> This makes sense. Having to pass the instance around is annoying, but unavoidable, and like you said there are ways to improve it in the future. And this will still be optional, you can still use the static if you want?

Yes, you can still use the statics if you want. Other than making `reinitalize` package private, none of the existing tests were changed.

> What I don't understand is if the instance remains a singleton, and what happens if you try to create multiple.

The instance is not a singleton on its own. I will add a test creating multiples

> Also we should probably do a quick test mocking the instance, to verify it works as we expect as that's one of the main use cases for this feature.

I'll add a test for this too

transact/src/main/java/dev/dbos/transact/execution/DBOSLifecycleListener.java
```diff
@@ -1,12 +1,14 @@
...
+import dev.dbos.transact.DBOS;
+
 /**
  * For registering callbacks that hear about `DBOS.launch()` and `DBOS.shutdown()`. At this point,
  * DBOS is ready to run workflows, and no additional registrations are allowed.
  */
 public interface DBOSLifecycleListener {
   /** Called from within DBOS.launch, after workflow processing is allowed */
-  void dbosLaunched();
+  void dbosLaunched(DBOS.Instance dbos);
```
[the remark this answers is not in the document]
devhawk: we're still in pre-release, so breaking changes are OK within reason

PR #295, opened 2026-03-03
PL/pgSQL enqueue & send

devhawk:
~~also moves migratons to resource files ~~

Also simplifies migration and removes migration error ignoring that was incorrectly ported from TS

### DIFF
--- a/transact/src/main/java/dev/dbos/transact/migrations/MigrationManager.java
+++ b/transact/src/main/java/dev/dbos/transact/migrations/MigrationManager.java
@@ -532,25 +532,28 @@ public class MigrationManager {
 
           -- Validate workflow metadata matches
           IF v_existing_name IS DISTINCT FROM workflow_name THEN
-              RAISE EXCEPTION 'DBOS_CONFLICTING_WORKFLOW: Workflow already exists with a different function name: %%s, but the provided function name is: %%s', v_existing_name, workflow_name;
+              RAISE EXCEPTION 'Conflicting DBOS workflow name'
+                 USING DETAIL = format('Workflow %%s exists with name %%s, but the provided workflow name is: %%s', v_workflow_id, v_existing_name, workflow_name),
+                      ERRCODE = 'invalid_parameter_value';
           END IF;
           IF v_existing_class_name IS DISTINCT FROM class_name THEN
-              RAISE EXCEPTION 'DBOS_CONFLICTING_WORKFLOW: Workflow already exists with a different class name: %%s, but the provided class name is: %%s', v_existing_class_name, class_name;
+              RAISE EXCEPTION 'Conflicting DBOS workflow class_name'
+                 USING DETAIL = format('Workflow %%s exists with class_name %%s, but the provided class_name is: %%s', v_workflow_id, v_existing_class_name, class_name),
+                      ERRCODE = 'invalid_parameter_value';
           END IF;
           IF v_existing_config_name IS DISTINCT FROM config_name THEN
-              RAISE EXCEPTION 'DBOS_CONFLICTING_WORKFLOW: Workflow already exists with a different class configuration: %%s, but the provided class configuration is: %%s', v_existing_config_name, config_name;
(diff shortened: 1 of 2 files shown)


transact/src/main/java/dev/dbos/transact/migrations/MigrationManager.java
```diff
@@ -233,232 +238,28 @@ static void runDbosMigrations(Connection conn, String schema, List<String> migra
...
-            migration8,
-            migration9,
-            migration10,
-            migration11,
-            migration12);
-    return migrations.stream().map(m -> m.formatted(schema)).toList();
+    List<String> migrations = new ArrayList<>();
+    int migrationNum = 1;
+    while (true) {
+      String resourceName = "/db/migrations/migration" + migrationNum + ".sql";
```
kraftp: Is this robust? I think "strings in memory" is safer even if it may be uglier. We don't have to have the strings in this file.

devhawk: I will go back to the previous approach

transact/src/main/java/dev/dbos/transact/migrations/MigrationManager.java
```diff
@@ -461,4 +463,137 @@ IF NOT EXISTS (
...
+      CREATE TABLE "%1$s".application_versions (
+        version_id TEXT NOT NULL PRIMARY KEY,
+        version_name TEXT NOT NULL UNIQUE,
+        version_timestamp BIGINT NOT NULL DEFAULT (EXTRACT(epoch FROM now()) * 1000.0)::bigint,
+        created_at BIGINT NOT NULL DEFAULT (EXTRACT(epoch FROM now()) * 1000.0)::bigint
+      );
+      """;
+
+  static final String migration14 =
+      """
```
kraftp, line 459: Let's say we wanted to change these--how would we do it?

For example, how would we add a new argument to `enqueue_workflow` (which we do fairly often)?

devhawk: AFAIK a future migration would have to drop the existing function then create a new one. If we aren't changing params, we could create or replace without needing to drop

kraftp: What worries me is that any change to the params--even adding new params--is breaking. And we're definitely going to add new params. Is there any way to avoid this?

devhawk: As long as the new parameters are at the end and have default values, previous invocation still works. Also, you can invoke PG functions with a mix of named & positional parameters, so you can do something like `SELECT dbos.enqueue_workflow(?, ?, ?, class_name => ?, deduplication_id => ?)`  without having to pass explicit NULLs for all the parameters between the first three (workflow name, queue name, positional parameters) and the ones you want to set.

Official docs: https://www.postgresql.org/docs/18/sql-syntax-calling-funcs.html

transact/src/test/java/dev/dbos/transact/migrations/BackCompatTest.java
```diff
@@ -45,10 +45,13 @@ void testInitialRun() throws Exception {
         Connection conn = dataSource.getConnection()) {
       DatabaseMetaData metaData = conn.getMetaData();
 
-      MigrationManagerTest.assertTableExists(metaData, "operation_outputs");
-      MigrationManagerTest.assertTableExists(metaData, "workflow_status");
-      MigrationManagerTest.assertTableExists(metaData, "notifications");
-      MigrationManagerTest.assertTableExists(metaData, "workflow_events");
+      for (var table : MigrationManagerTest.EXPECTED_TABLES) {
```
kraftp: I think the entire test file is redundant and can now be removed?

devhawk: testInitialRun is redunant, testWayFutureVersion is not. But I can move it into the MigrationManagerTest file

PR #309, opened 2026-03-11
DBOS Instance

### LINKED ISSUE #306: Remove DBOS static API in favor of instance based API
Integrating DBOS into applications that use dependency injection like Spring Boot is problematic due to the static method focus DBOS has today. In order to better integrate into these apps, we are going to make DBOS Java purely instance based and do away with the static APIs

devhawk:
This PR is a **_major breaking change_** to the DBOS public API surface area. We are removing the global singleton and most of the static methods on `DBOS` in favor of using DBOS as an instance. We had already done the work to ensure that everything worked with `DBOS.Instance` directly. However, the combination of static and instance APIs made for a poor developer experience, so we've decided to remove the static API entirely. 

Mostly, this PR is test work since nearly all the tests used the static API. Updating all the tests allowed us to make a couple other changes - in particular, using [`@AutoClose`](https://docs.junit.org/6.0.3/writing-tests/built-in-extensions.html#AutoClose) to simplify test cleanup and using [TestContainers](https://testcontainers.com/) for database isolation.  The combination of DBOS Instance and TestContainers enables us to run tests in parallel, bringing a test run down to 6 minutes from 15.

> Major thanks to @hannosgit who did the original TestContainers work!

In addition to the Java work, this PR updates the Kotlin DBOSExtension class. The previous DBOSExtension primarily provided top level Kotlin functions for DBOS static methods. Now, the DBOSExtension simply wraps `DBOS.startWorkflow` and `DBOS.runStep` to support [Kotlin trailing lambdas syntax](https://kotlinlang.org/docs/lambdas.html#passing-trailing-lambdas). We also added a Kotlin test!

fixes #306

### DIFF
--- a/transact/src/main/java/dev/dbos/transact/DBOS.java
+++ b/transact/src/main/java/dev/dbos/transact/DBOS.java
@@ -174,64 +174,58 @@ public class DBOS implements AutoCloseable {
         interfaceClass, implementation, instanceName, () -> this.dbosExecutor.get());
   }
 
-  private void registerClassWorkflows(Class<?> interfaceClass, Object target, String instanceName) {
-    if (dbosExecutor.get() != null) {
-      throw new IllegalStateException("Cannot register workflow after DBOS is launched");
-    }
+  private void registerClassWorkflows(
+      @NonNull Class<?> interfaceClass,
+      @NonNull Object implementation,
+      @Nullable String instanceName) {
     Objects.requireNonNull(interfaceClass, "interfaceClass must not be null");
+    Objects.requireNonNull(implementation, "implementation must not be null");
     instanceName = Objects.requireNonNullElse(instanceName, "");
     if (!interfaceClass.isInterface()) {
       throw new IllegalArgumentException("interfaceClass must be an interface");
     }
-    Objects.requireNonNull(target, "implementation must not be null");
-
-    // Use @WorkflowClassName annotation if present, otherwise use the Java class name
-    WorkflowClassName classNameAnnotation =
-        target.getClass().getAnnotation(WorkflowClassName.class);
-    String className =
-        (classNameAnnotation != null && !classNameAnnotation.value().isEmpty())
-            ? classNameAnnotation.value()
-            : target.getClass().getName();
-    workflowRegistry.register(interfaceClass, target, className, instanceName);
-
-    Method[] methods = target.getClass().getDeclaredMethods();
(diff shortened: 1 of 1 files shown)


TEST_REFACTOR_PLAN.md
```diff
@@ -0,0 +1,165 @@
+# Test Refactoring Plan: Migrating to `DBOS.Instance` and `PgContainer`
```
kraftp: Let's not commit this

transact/src/main/kotlin/dev/dbos/transact/DBOSExtensions.kt
```diff
@@ -5,166 +5,107 @@ package dev.dbos.transact
...
-@JvmSynthetic
-fun <T> startWorkflow(options: StartWorkflowOptions, block: () -> T): WorkflowHandle<T, Exception> {
-    return DBOS.startWorkflow(ThrowingSupplier<T, Exception> { block() }, options)
-}
-
-@JvmSynthetic
-@JvmName("startWorkflowUnit")
-fun startWorkflow(block: () -> Unit): WorkflowHandle<Void, Exception> {
-    return DBOS.startWorkflow(ThrowingRunnable<Exception> { block() })
+fun <T> DBOS.beginWorkflow(
```
kraftp: Is this needed? We could instead always require options to be passed in Kotlin so there aren't two mostly-redundant core API methods.

TEST_REFACTOR_PLAN.md
devhawk: agreed

transact/src/main/kotlin/dev/dbos/transact/DBOSExtensions.kt
devhawk: The issue is that we can't hide the one parameter version of startWorkflow from Kotlin. And trying to use it from Kotlin is a pretty poor dev experience. Unlike the two param overload or `runStep`, you have to specify the generic types explicitly from Kotlin:

```kt
dbos.startWorkflow<Unit, Exception> { 
  proxy.exampleWorkflow() 
}
```

And if you don't specify the generic types, you get the less-than-useful error `Cannot infer type for type parameter 'E'. Specify it explicitly.`. 

I think we need a better solution here, but I want to give it more thought, so in the meantime I'll remove the `beginWorkflow` wrappers.

PR #327, opened 2026-03-25
App Version Support

### LINKED ISSUE #314: Application Version Management
Analogous to https://github.com/dbos-inc/dbos-transact-py/pull/598

devhawk:
fixes #314

### DIFF
--- a/transact/src/test/java/dev/dbos/transact/client/ClientTest.java
+++ b/transact/src/test/java/dev/dbos/transact/client/ClientTest.java
@@ -150,7 +150,7 @@ public class ClientTest {
     }
   }
 
-  @Test
+  @RetryingTest(3)
   public void clientListApplicationVersions() throws Exception {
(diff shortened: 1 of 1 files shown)


transact/src/test/java/dev/dbos/transact/workflow/WorkflowMgmtTest.java
```diff
@@ -193,6 +194,56 @@ public void testListWorkflowsDontLoadInputOutput() throws Exception {
     assertNull(workflow.serialization());
   }
 
+  @Test
+  public void testListApplicationVersions() throws Exception {
```
kraftp, line 198: Can we update these tests to work like those in Python/TS, where it repeatedly launches and destroys the DBOS instance and verifies appropriate versions are created and workflows (including scheduled workflows) have those versions?

devhawk: can do

PR #348, opened 2026-04-10
DELAYED support

devhawk:
this PR adds support for the new `DELAYED` wf state, which is liked `ENQUEUED` but with a specified dequeue time typically in the future.

Also adds `withDelay` for `StartWorkflowOptions` and `DBOSClient.EnqueueOptions` as well as `setWorkflowDelay` for `DBOS` and `DBOSClient`. `QueueService` calls `SystemDatabase.transitionDelayedWorkflows` every second to move `DELAYED` workflows that have reached their delay until value to `ENQUEUED`

Other changes
* `insertWorkflowStatus` no longer accepts state input, it's generated based on queue name & delay
* `listWorkflows` queues includes `DELAYED`
* `garbageCollect` does not delete `DELAYED`
* `insertWorkflowStatus` doesn't update recovery attempts or executor id for `DELAYED` in addition to `ENQUEUED`
* `StepInfo` and `NotificationInfo` now have strongly typed temporal values
* moved timeout+deadline conflict check to `DBOSContext.setContext` and `DBOSExecutor.startWorkflow`
* moved queue partition key + dedupe ID conflict check to `DBOSExecutor.enqueueWorkflow`


transact/src/main/java/dev/dbos/transact/execution/QueueService.java
```diff
@@ -77,6 +77,11 @@ private void startQueueListeners(List<Queue> queues, Set<String> listenQueues) {
     final Duration minPollingInterval = Duration.ofSeconds(1);
     final Duration maxPollingInterval = Duration.ofSeconds(120);
 
+    var execService = execServiceRef.get();
+    if (execService != null) {
+      execService.scheduleAtFixedRate(this::transitionDelayedWorkflows, 1, 1, TimeUnit.SECONDS);
+    }
```
maxdml, line 83: Wouldn't it be simpler to do that in the same scheduled job than the dequeue logic?

maxdml (commented):
I might have missed it, but did you make import/export workflows handle the delay field?

transact/src/main/java/dev/dbos/transact/execution/QueueService.java
devhawk: I didn't think so. In Java, we create a `Runnable` per queue to poll for workflows to dequeue. `transitionDelayedWorkflows` isn't queue specific, so I thought it was cleaner to have its own `Runnable`. I mean, it's one line of code here and `this::transitionDelayedWorkflows` is just a call to `systemDatabase.transitionDelayedWorkflows` inside a try catch block (which is needed because throwing inside a scheduled runnable kills future scheduling. Not sure how much simpler it can get

devhawk:
> I might have missed it, but did you make import/export workflows handle the delay field?

No, I missed it. Good catch.

maxdml:
@devhawk also needs to filter out delayed workflows in the global timeout logic:

```
  public void globalTimeout(Instant endTime) {
    var input = new ListWorkflowsInput().withEndTime(endTime);
    for (WorkflowStatus status :
        systemDatabase.listWorkflows(input.withStatus(WorkflowState.PENDING))) {
      cancelWorkflows(List.of(status.workflowId()));
    }

    for (WorkflowStatus status :
        systemDatabase.listWorkflows(input.withStatus(WorkflowState.ENQUEUED))) {
      cancelWorkflows(List.of(status.workflowId()));
    }
  }
```

devhawk:
> @devhawk also needs to filter out delayed workflows in the global timeout logic:

fixed

transact/src/main/java/dev/dbos/transact/conductor/protocol/ExportWorkflowRequest.java
```diff
@@ -7,7 +7,7 @@ public class ExportWorkflowRequest extends BaseMessage {
   public ExportWorkflowRequest() {}
 
   public ExportWorkflowRequest(String requestId, String workflowId, boolean exportChildren) {
-    this.type = MessageType.DELETE.getValue();
+    this.type = MessageType.EXPORT_WORKFLOW.getValue();
```
maxdml, line 10: wut! lol

maxdml (approved):

transact/src/main/java/dev/dbos/transact/conductor/protocol/ExportWorkflowRequest.java
devhawk: The request class constructors are never used, which is why this didn't get caught. It was only test code that ever used it but then I switched the test code to use raw JSON so I could set additional fields for test purposes.

Issue #349, opened 2026-04-13
Xplat Create Schedules - standard cron


devhawk:
Not sure if this is a significant issue, but if there is variance in the cron syntax between languages (lang a client creates a schedule in lang b dbos) will we run into a scenario where the cron is not parsable by the target language. Is this an issue?

PR #350, opened 2026-04-13
Improve WF Schedule API

devhawk:
Pass a WorkflowSchedule object to createSchedule instead of a bunch of parameters

### DIFF
--- a/transact/src/main/java/dev/dbos/transact/DBOS.java
+++ b/transact/src/main/java/dev/dbos/transact/DBOS.java
@@ -32,7 +32,6 @@ import java.io.InputStream;
 import java.lang.reflect.Method;
 import java.time.Duration;
 import java.time.Instant;
-import java.time.ZoneId;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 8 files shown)


devhawk:
Note @kraftp this PR moves the cron string validation into `SystemDatabse.createSchedule` which could potentially lead to lang portability issues. Tracked by #349. Let me know if you want me to pull cron validation back to `DBOSExecutor` where it would only be validated for intra-language scheduling. 

The [cron parsing library](https://github.com/jmrozanec/cron-utils) we are currently using supports [multiple cron syntaxes](https://github.com/jmrozanec/cron-utils#:~:text=Pre%2Ddefined%20definitions) which implies there isn't a single standard cron syntax. 

Maybe we should port the [TS cron parser](https://github.com/dbos-inc/dbos-transact-ts/blob/main/src/scheduler/crontab.ts) over?

kraftp:
> Note @kraftp this PR moves the cron string validation into `SystemDatabse.createSchedule` which could potentially lead to lang portability issues. Tracked by #349. Let me know if you want me to pull cron validation back to `DBOSExecutor` where it would only be validated for intra-language scheduling.
> 
> The [cron parsing library](https://github.com/jmrozanec/cron-utils) we are currently using supports [multiple cron syntaxes](https://github.com/jmrozanec/cron-utils#:~:text=Pre%2Ddefined%20definitions) which implies there isn't a single standard cron syntax.
> 
> Maybe we should port the [TS cron parser](https://github.com/dbos-inc/dbos-transact-ts/blob/main/src/scheduler/crontab.ts) over?

Not worried about cron syntax compatibility across languages. There just isn't a unified enough standard, and we shouldn't be writing our own parsers. TS and Python are already incompatible. We should just pick a good Java standard and support it well.

kraftp (approved):
This definitely looks better!

devhawk:
> Not worried about cron syntax compatibility across languages. There just isn't a unified enough standard, and we shouldn't be writing our own parsers. TS and Python are already incompatible. We should just pick a good Java standard and support it well.

Make sense, tho it's kind of weird that we have custom cron parsers in TS and Python but they aren't compatible?

PR #352, opened 2026-04-14
fix recent test failures

devhawk:


devhawk:
Note, since this is just test fixes I went ahead and merged w/o review

PR #356, opened 2026-04-17
transact-spring-boot-starter

### LINKED ISSUE #145: Integration with Spring
Can we integrate DBOS with Spring or is it meant to be independent of it? I see that we have to define the DB config again and I understand that this approach is meant for it to not rely entirely on the spring framework, but can we avoid it?

devhawk:
This PR adds SpringBoot integration for DBOS. While we will add deeper integration later, this PR adds:

Updated version of WidgetStore demo app using the code from this PR can be seen here: https://github.com/devhawk/widget-store-java. The main demo app repo will be updated with this version of WidgetStore when the spring boot integration is officially released.

Changes:
* support for dbos.* SpringBoot properties. 
  * use spring application.name and datasource properties as fall back if dbos versions are not set
* auto configure DBOSConfig and DBOS beans
* auto configures DBOSLifecycle bean to launch/shutdown dbos on spring boot lifecycle events
* auto configures DBOSWorkflowRegistrar to automatically register all singleton bean @Workflow methods with DBOS
* auto configures DBOSAspect bean to automatically proxy @Workflow/@Step calls to invoke runWorkflow / runStep 
* exposed runWorkflow and moved startRegisteredWorkflow from DBOS -> DBOSIntegration. This way they are still accessable but not part of the primary API surface area.

fixes #145


kraftp (commented):
This is pretty cool!  So looking at the example code, the only place DBOS is needed is in `WidgetStoreService.java` to actually define the workflows and such.

If DBOS is auto-configured, how do I manually configure it (set the database URL, etc.) if I want to?

devhawk:
> This is pretty cool! So looking at the example code, the only place DBOS is needed is in `WidgetStoreService.java` to actually define the workflows and such.
> 
> If DBOS is auto-configured, how do I manually configure it (set the database URL, etc.) if I want to?

You have multiple options: 

* First, you can set `DBOSConfig` fields via `dbos.*` spring boot properties (see `DBOSProperties` class). We only use the spring app name and database source properties if equivalent dbos properties aren't set.
* Next, you can define one or more `DBOSConfigCustomizer` instances to modify the `DBOSConfig` instance before `DBOS` instance is created
* Finally, both `DBOSConfig` and `DBOS` bean methods are annotated `@ConditionalOnMissingBean`. If you define your own bean construction methods for either or both of those instances, the library provided versions from `DBOSAutoConfiguration` will not be used.

kraftp (approved):

PR #358, opened 2026-04-20
Varargs version of applySchedules

devhawk:
also enforce settings for created schedules (new id, active status, null last fired at) and added more tests


kraftp (approved):
New syntax looks good!

PR #363, opened 2026-04-22
Update AppVersionComputer to use registered workflow FQ name & method's byte code

### LINKED ISSUE #328: revisit app version computer
we should be hashing on registered workflows not classes

devhawk:
fixes #328

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,4 +1,5 @@
 [versions]
+asm = "9.7.1"
 aspectj = "1.9.22.1"
 assertj = "3.27.3"
 cron-utils = "9.2.1"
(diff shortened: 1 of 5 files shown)


kraftp (commented):
This is fairly complex--are we confident this is stable across all environments where Java may be run? Otherwise we can consider a simpler approach and recommend users either version manually or use patching.

devhawk:
> This is fairly complex--are we confident this is stable across all environments where Java may be run? Otherwise we can consider a simpler approach and recommend users either version manually or use patching.

We should definitely recommend users version manually or use patching. Hasn't calculating app version from code always been a fallback scenario?

Given that the Java class file format is standardized, I'm not worried about the complexity. JDK itself used ASM library for class file parsing and byte code generation until recently (they added a [class file API](https://openjdk.org/jeps/484) in JDK 22 and finalized it in JDK 24). 

AoT compilation breaks app version calculation, but it would break both the current version as well as the new version introduced by this PR.

kraftp (approved):

PR #370, opened 2026-04-24
Update JSON handling

### LINKED ISSUE #369: Review JSON Handling
Breakup JSONUtil ball of mud

devhawk:
This PR breaks up the JSONUtil [big ball of mud](https://en.wikipedia.org/wiki/Spaghetti_code#Big_ball_of_mud). Instead of a single ObjectMapper instance shared by different parts of the code with different needs, this has no been broken up as needed for different parts of the code:

* `JsonUtility` has a static "vanilla" object mapper that is used both by `DBOSPortableSerializer` as well as places in the code where we need generic to/from JSON support (admin server & conductor protocol, authenticatedRoles string array)
* `DBOSJavaSerializer` includes an object mapper with a custom type resolver. This enables java type info to be included in the serialzied JSON for accurate deserialization.
* 'Conductor` has an object mapper configured for large strings which may be recived as import requests

Additionally, the custom type resolver approach in `DBOSJavaSerializer` eliminates the need for boxing or wrapping values in array as `JSONUtil` did. This allowed us to remove the `noHistoricalWrapper` parameter from DBOSSerializer methods. Since this is a breaking change for DBOSSerializer (and because there are other breaking changes in the next DBOS Java release) we also changed the method names from javascripty `stringify` & `parse` to `serialize` & `deserialize`.

fixes #369


gradle/libs.versions.toml
```diff
@@ -4,6 +4,7 @@ aspectj = "1.9.22.1"
 assertj = "3.27.3"
 cron-utils = "9.2.1"
 hikaricp = "7.0.2"
+kryo = "5.6.2"
```
kraftp, line 7: Not sure I understand this--are these dependencies or dev dependencies? Is Kryo just used in tests or is it used in the actual serializer?

devhawk: gradle/libs.versions.toml is just the centralized list of versions to ensure we use the same version of dependencies in all our packages. You still have to specify the dependencies in build.gradle.kts files. Kyro is used in tests. You can see the `testImplementation` declaration here: https://github.com/dbos-inc/dbos-transact-java/blob/devhawk/fix-json/transact/build.gradle.kts#L49

kraftp (commented):
Just for validation, have we run any interop tests with other languages (not sure if there are any in CI)?

devhawk:
> Just for validation, have we run any interop tests with other languages (not sure if there are any in CI)?

I don't think any of our languages do full interop tests. None of the other GH test workflows install any additional language runtimes. I will manually run the dbos cloud `transact-interops` tests w/ this PR + make sure we have sufficient interop tests in this repo

devhawk:
@kraftp Verified locally that this PR passes all the `transact-interop` tests in the cloud repo

devhawk:
Also @kraftp, PortableSerializationTest.java tests a variety of portable interop scenarios.

kraftp (approved):
Looks good!

PR #374, opened 2026-04-29
Build and test demo apps from PR GH action

devhawk:


kraftp (approved):
This is nice, but may become a pain if we ever actually want to make breaking changes again

devhawk:
> This is nice, but may become a pain if we ever actually want to make breaking changes again

That's the whole point - to catch breaking changes!

PR #382, opened 2026-05-12
Update WF registration API

### LINKED ISSUE #381: improve DBOSIntegration API
* need a version of `registerWorkflow` that takes `@Workflow` values as parameters 
* `registerWorkflow` should return the `RegisteredWorkflow` instance
* `getRegisteredWorkflow`/`getRegisteredWorkflowInstances`/`getRegisteredWorkflows` should all work pre launch

devhawk:
* Adds DBOSIntegration.registerWorkflow overload that takes direct parameters instead of @Workflow annotation instance
* **BREAKING CHANGE**: changed DBOSIntegration.registerWorkflow overloads to return the registered workflow
* Modifled DBOSIntegration getRegisteredWorkflows/getRegisteredWorkflow/getRegisteredWorkflowInstances to work pre and post launch
* **BREAKING CHANGE**: removed DBOS.registerWorkflow method. Call DBOSIntegration.registerWorkflow instead

fixes #381

### DIFF
--- a/transact/src/main/java/dev/dbos/transact/DBOS.java
+++ b/transact/src/main/java/dev/dbos/transact/DBOS.java
@@ -4,6 +4,7 @@ import dev.dbos.transact.config.DBOSConfig;
 import dev.dbos.transact.context.DBOSContext;
 import dev.dbos.transact.execution.DBOSExecutor;
 import dev.dbos.transact.execution.DBOSLifecycleListener;
+import dev.dbos.transact.execution.RegisteredWorkflow;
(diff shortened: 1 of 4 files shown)


kraftp (approved):
This only affects integrations, right, not normal users? Looks good!

devhawk:
> This only affects integrations, right, not normal users? Looks good!

Yes. I made the breaking change to move `registerWorkflow` from `DBOS` to `DBOSIntegration` because it's not a "normal user" API

PR #383, opened 2026-05-12
SysDB Static DAOs

devhawk:
Given the complexity of SystemDatabase, I've come to ["see the light"](https://www.youtube.com/watch?v=aCKSz1N6fpQ) on grouping related code into DAOs. However, I'm not a fan of DAO instances, so this PR does two things:

* Moves db logic that had been added to SystemDatabase directly to DAOs
* Makes all the DAOs purely containers for static methods (in C#, they'd be "static classes" but Java has no such construct). 
  * All the DAOs have private constructors to ensure we don't create instances of them

> Note, NotificationService remains a separate non-static class since it does hold resources specific to notification listening (thread reference and notification map)


kraftp (approved):
No functional changes, looks good

PR #388, opened 2026-05-18
CockroachDB Support

devhawk:
This PR adds CockroachDB support. It also generally revamps how SysDB is organized and how Listen/Notify events are delivered.

Changes:
* All DAO classes moved into database/dao directory
* Added DbContext to carry data source, schema, serializer and a closed check in a single DAO parameter
* Added new SysDB signal infrastructure (SignalMap, SignalKey, Subscription)
* Rewrote recv and getEvent to use new signal infrastructure 
* Added DBOSConfig useListenNotify field
* Added cockroach DB check in sysdb ctor to override useListenNotify value
* renamed NotificationService -> NotificationListenerSource and reworked to use new signal infrastructure
* added NotifcationRegistry and NotificationSource interfaces 
* added NullNotificationSource to use when useListenNotify is false or running on CockroachDB
* Updated migration manager to split migration 1 into two parts (listen/notify trigger + function and every thing else) and move migration 10 check to java for CRDB compat
* reworked test infrastructure to truncate DBOS tables rather than dropping the DB to save on migration execution time (very very very slow on CRDB)
* added DBOS_TEST_USE_COCKROACH_DB env var to run test suite on CRDB instead of PG
* Added ParallelExecutionConfigurationStrategy implementation to use fixed config for CRDB since those containers are much more resource intensive than the PG containers


.github/workflows/on_pr.yml
```diff
@@ -14,5 +14,7 @@ on:
 jobs:
   test:
     uses: ./.github/workflows/test.yml
+  test-crdb:
```
kraftp, line 17: These take a little bit longer than the regular tests--do we want them blocking PRs or should they be in the background?

transact/src/test/java/dev/dbos/transact/utils/PgContainer.java
```diff
@@ -3,85 +3,112 @@
...
-  }
-
-  static PostgreSQLContainer acquire() {
-    try {
-      PERMITS.acquire();
-      var container = POOL.poll();
-      if (container == null) {
-        container = new PostgreSQLContainer("postgres:18");
-        container.start();
+  public static final boolean USE_COCKROACH_DB =
```
kraftp, line 22: It's awesome that we have all tests running on CRDB!

.github/workflows/on_pr.yml
devhawk: we run .github/workflows/test.yml, .github/workflows/test_crdb.yml and .github/workflows/test_demo_apps.yml on every PR and main Push, but only the test.yml tests are required for merging. the crdb and demo apps test runs don't block PRs

kraftp (approved):

PR #389, opened 2026-05-18
Update Migrations

devhawk:
* Adds Migrations 20-35
* restructures migration manager to run each migration (except concurrent ones) in a single transaction
* removes test-crdb from on push (it's in on-pr, doesn't need to be in both)


transact/src/test/java/dev/dbos/transact/migrations/MigrationManagerTest.java
```diff
@@ -88,7 +88,10 @@ void testRunMigrations_CreatesTables() throws Exception {
         }
       }
 
-      var migrations = new ArrayList<>(MigrationManager.getMigrations(Constants.DB_SCHEMA, true));
+      var migrations =
+          new ArrayList<>(
```
kraftp, line 93: We should probably add the additional tests from Python for handling invalid indexes, online migration idempotence, etc.

kraftp (approved):

PR #390, opened 2026-05-19
Add Debouncer — coalesce rapid workflow calls into one execution

### LINKED ISSUE #288: Debouncing support

easmith:
Implements a debounce mechanism for DBOS workflows analogous to dbos-transact-py _debouncer.py. Multiple calls with the same key within a period are collapsed into a single user-workflow execution that runs with the most recently supplied arguments.

Architecture:
- DebouncerServiceImpl: internal @Workflow that runs a recv-loop, absorbing messages until the debounce period times out or the absolute debounceTimeout elapses, then starts the user workflow.
- Debouncer<R>: public fluent API. Enqueues the service workflow on _dbos_internal_queue with a deduplicationId derived from (workflowName, debounceKey). On DBOSQueueDuplicatedException, forwards a message to the running debouncer and waits for an ack.
- DBOSExecutor.captureInvocation(): extracted from startWorkflow so Debouncer can capture a lambda's workflow call without executing it.
- Auto-registration of DebouncerService in DBOS constructor so users need no boilerplate setup.
- Internal system workflows filtered from getRegisteredWorkflows / getRegisteredWorkflowInstances to keep public counts clean.

Usage:

```java
var handle = dbos.<String>debouncer()
    .withDebounceTimeout(Duration.ofMinutes(5))
     .debounce("key", Duration.ofSeconds(2), () -> svc.process(arg)); 
String result = handle.getResult();
```
Tests: 6 integration tests via Testcontainers Postgres covering single-call, multi-call coalescing, absolute timeout, independent keys, concurrent callers, and queue-based user workflow.


devhawk:
Hey @easmith, thanks for the contribution! I was traveling today so I haven't had a chance to review this, but I kicked off the GH actions run. I'll take a look tomorrow. (USA Pacific time zone)

transact/src/main/java/dev/dbos/transact/workflow/internal/InternalWorkflows.java
```diff
@@ -0,0 +1,126 @@
...
+
+  @Override
+  @Workflow(name = Constants.DEBOUNCER_WORKFLOW_NAME)
+  public void debouncerWorkflow(
+      DebouncerOptions options, DebouncerContextOptions ctx, DebouncerMessage initial) {
+
+    // Publish the pre-assigned user workflow id as an event so callers waiting on the
+    // deduplication path can retrieve it via getEvent without relying on Jackson deserializing
+    // workflow inputs (records are final classes so @class type info is not emitted).
+    dbos.setEvent(Constants.DEBOUNCER_CHILD_ID_KEY, ctx.userWorkflowId());
```
devhawk, line 57: I'm cool with setting this as an event, but note that the comment is incorrect. DBOSJavaSerializer always includes type information (except for "natural" JSON types Double, Integer, Boolean and String). Records are fine to use as workflow inputs. See JavaSerializerTest for examples of serializing records

transact/src/main/java/dev/dbos/transact/workflow/internal/InternalWorkflows.java
```diff
@@ -0,0 +1,126 @@
...
+
+    // Record the absolute deadline once as a durable step. On recovery this returns the same
+    // value so the loop's exit condition is replay-stable across crashes.
+    long deadlineEpochMs =
```
devhawk, line 61: Use strongly typed Duration/Instant classes for time related calculations

transact/src/main/java/dev/dbos/transact/workflow/internal/DebouncerServiceImpl.java
```diff
@@ -0,0 +1,126 @@
...
+      // Acknowledge receipt so the sender knows the message was consumed by this loop iteration.
+      dbos.setEvent(next.messageId(), next.messageId());
+    }
+
+    var optWorkflow =
+        dbos.integration()
+            .getRegisteredWorkflow(
+                options.workflowName(),
+                options.className(),
+                options.instanceName() == null ? "" : options.instanceName());
```
devhawk: Looks like I have a bug in getRegisteredWorkflow nullability annotations. Instance name should be @Nullable. I'll need to test, but the instance name param of RegisteredWorkflow.fullyQualifiedName is @Nullable

transact/src/main/java/dev/dbos/transact/workflow/internal/DebouncerOptions.java
```diff
@@ -0,0 +1,17 @@
...
+ * and the optional absolute timeout cap.
+ *
+ * <p>Not part of the public API.
+ */
+public record DebouncerOptions(
+    @NonNull String workflowName,
+    @NonNull String className,
+    @Nullable String instanceName,
+    @Nullable String queueName,
+    @Nullable Long debounceTimeoutMs) {}
```
devhawk: Use strongly typed time classes (Duration not Long)

transact/src/main/java/dev/dbos/transact/workflow/internal/DebouncerMessage.java
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Message sent from a {@code Debouncer} caller to the debouncer service workflow each time the
+ * debounce key fires. The debouncer service workflow uses the most recently received message's args
+ * when it eventually starts the user workflow.
+ *
+ * <p>Not part of the public API — the debouncer infrastructure consumes this directly.
+ */
+public record DebouncerMessage(
+    @NonNull String messageId, @NonNull Object[] args, long debouncePeriodMs) {}
```
devhawk: Use strongly typed time classes (Duration not Long)

transact/src/main/java/dev/dbos/transact/workflow/internal/DebouncerContextOptions.java
```diff
@@ -0,0 +1,18 @@
...
+ * so that the caller can return a handle pointing to the future workflow.
+ *
+ * <p>Not part of the public API.
+ */
+public record DebouncerContextOptions(
+    @NonNull String userWorkflowId,
+    @Nullable String deduplicationId,
+    @Nullable Integer priority,
+    @Nullable String appVersion,
+    @Nullable Long workflowTimeoutMs) {}
```
devhawk: Use strongly typed time classes (Duration not Long)

transact/src/main/java/dev/dbos/transact/workflow/internal/DebouncerServiceImpl.java
```diff
@@ -0,0 +1,126 @@
...
+            .withQueue(options.queueName())
+            .withDeduplicationId(hasQueue ? ctx.deduplicationId() : null)
+            .withPriority(hasQueue ? ctx.priority() : null)
+            .withAppVersion(ctx.appVersion());
+    if (ctx.workflowTimeoutMs() != null) {
+      startOpts = startOpts.withTimeout(Duration.ofMillis(ctx.workflowTimeoutMs()));
+    }
+
+    // Coerce args to the method's declared parameter types before invocation.
+    // Jackson's type-info round-trip through send/recv (Object[]) can produce numeric
```
devhawk: Have you actually seen this? See comment above about DBOSJavaSerializer configured to always write class information to avoid exactly what you describe here

transact/src/main/java/dev/dbos/transact/workflow/internal/DebouncerServiceImpl.java
```diff
@@ -0,0 +1,126 @@
...
+    // mismatches (e.g. long → Integer) that cause IllegalArgumentException at reflection
+    // call-site. This mirrors the coercion already applied in executeWorkflowById.
+    var workflow = optWorkflow.orElseThrow();
```
devhawk: You can combine the earlier `optWorkflow.isEmpty` call with the Supplier overload of orElseThrow

transact/src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -84,6 +88,9 @@ public DBOS(@NonNull DBOSConfig config) {
     this.integration =
         new DBOSIntegration(
             this.config, this.workflowRegistry, dbosExecutor::get, this::registerLifecycleListener);
+    // Register the built-in debouncer service workflow so callers can use Debouncer without
+    // having to declare and wire the service themselves.
+    this.debouncerProxy = registerProxy(DebouncerService.class, new DebouncerServiceImpl(this));
```
devhawk: use `workflowRegistry.registerWorkflow` to register the internal workflow method instead of using the proxy. Pass the registered workflow to debouncer and use `startRegisteredWorkflow`

transact/src/main/java/dev/dbos/transact/internal/DBOSIntegration.java
```diff
@@ -153,6 +156,53 @@ public RegisteredWorkflow registerWorkflow(
...
+   * Capture the workflow invocation triggered by the supplied lambda without executing the
+   * workflow. The lambda must call exactly one {@code @Workflow} method on a registered proxy. This
+   * is intended for infrastructure that needs to defer a workflow start (for example, the
+   * debouncer).
+   *
+   * @param wfLambda lambda that invokes exactly one workflow method on a registered proxy
+   * @return the captured workflow name, class name, instance name, and arguments
+   * @throws IllegalStateException if DBOS has not been launched
+   */
+  public <T, E extends Exception> CapturedInvocation captureInvocation(
```
devhawk: I generally don't like to introduce new public APIs that aren't needed. Since Debouncer is a built in feature, it could take an executor instance and call `captureInvocation` and `findWorkflowIdByDeduplicationId` directly

transact/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
```diff
@@ -1205,6 +1209,36 @@ public Object runWorkflow(
     }
   }
 
+  /**
+   * Capture the workflow invocation triggered by the supplied lambda without executing the
+   * workflow. The lambda must call exactly one @Workflow method on a registered proxy on this
+   * executor.
+   */
+  public <T, E extends Exception> Invocation captureInvocation(ThrowingSupplier<T, E> wfLambda) {
```
devhawk, line 1312: nice change.

transact/src/main/java/dev/dbos/transact/workflow/Debouncer.java
```diff
@@ -0,0 +1,270 @@
...
+    this.queueName = queueName;
+    this.debounceTimeout = debounceTimeout;
+  }
+
+  /**
+   * Set the queue that the user workflow will be enqueued on when the debounce period elapses.
+   * {@code null} starts the user workflow directly (not enqueued).
+   */
+  public @NonNull Debouncer<R> withQueue(@Nullable String queueName) {
+    return new Debouncer<>(dbos, debouncerProxy, queueName, debounceTimeout);
```
devhawk: Add check to ensure queue name can be null but not empty

transact/src/main/java/dev/dbos/transact/workflow/Debouncer.java
```diff
@@ -0,0 +1,270 @@
...
+  /**
+   * Debounce a workflow with no return value. The supplier's return value is ignored — pass {@code
+   * null} when no value is available.
+   *
+   * @param debounceKey key that groups concurrent calls; calls with the same key are coalesced
+   * @param debouncePeriod inactivity window before the user workflow runs; each call resets it
+   * @param wfLambda lambda calling exactly one {@code @Workflow} method
+   * @return handle to the future user workflow
+   */
+  public @NonNull <E extends Exception> WorkflowHandle<Void, E> debounceVoid(
```
devhawk: this should be named just `debounce` like the version that takes ThrowingSupplier so it matches the pattern of `DBOS.runStep`

transact/src/main/java/dev/dbos/transact/workflow/Debouncer.java
```diff
@@ -0,0 +1,270 @@
...
+      throw new IllegalArgumentException("debouncePeriod must be a positive non-zero duration");
+    }
+
+    DBOSIntegration.CapturedInvocation invocation = dbos.integration().captureInvocation(wfLambda);
+
+    // When called from inside a workflow, UUID generation must be wrapped in a durable step so
+    // the same IDs are produced on replay. UUIDs are joined with "|" (not present in UUID format)
+    // to avoid two separate step increments.
+    String ids;
+    if (DBOS.inWorkflow() && !DBOS.inStep()) {
```
devhawk: If we had an executor reference, we could just call `runDbosFunctionAsStep` here

transact/src/main/java/dev/dbos/transact/workflow/Debouncer.java
```diff
@@ -0,0 +1,270 @@
...
+      ids = dbos.runStep(() -> UUID.randomUUID() + "|" + UUID.randomUUID(), "assignDebounceIds");
```
devhawk: One of these should be `DBOSContext.getNextWorkflowId`. Also, create a local record class to return the two values as a single struct to avoid having to serialize and then parse the strings

transact/src/main/java/dev/dbos/transact/workflow/Debouncer.java
```diff
@@ -0,0 +1,270 @@
...
+            invocation.className(),
+            invocation.instanceName(),
+            queueName,
+            debounceTimeout == null ? null : debounceTimeout.toMillis());
+    // Propagate the calling workflow's context (priority, timeout, appVersion, deduplicationId)
+    // to the user workflow — mirrors Python's ContextOptions snapshot.
+    Long timeoutMs = null;
+    if (DBOS.inWorkflow()) {
+      var wfCtx = DBOSContextHolder.get();
+      if (wfCtx != null && wfCtx.getTimeout() != null) {
```
devhawk: `DBOSContextHolder.get` never returns null

transact/src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -389,6 +396,21 @@ public void sleep(@NonNull Duration duration) {
...
+   * execution of the targeted workflow using the most recent arguments.
+   *
+   * <p>The returned debouncer is immutable; configuration helpers like {@link
+   * Debouncer#withQueue(String)} and {@link Debouncer#withDebounceTimeout(java.time.Duration)}
+   * return new instances.
+   *
+   * @param <R> the return type of the debounced workflow (used only for type inference)
+   * @return a fresh debouncer bound to this DBOS instance
+   */
+  public <R> @NonNull Debouncer<R> debouncer() {
```
devhawk, line 492: We also need a DBOSClient version of debouncer

transact/src/main/java/dev/dbos/transact/workflow/Debouncer.java
```diff
@@ -0,0 +1,270 @@
...
+    // the notifications table until processed, so we must not re-send on each ack-timeout retry.
+    boolean messageSent = false;
+    while (true) {
+      try {
+        var startOpts =
+            new StartWorkflowOptions()
+                .withQueue(Constants.DBOS_INTERNAL_QUEUE)
+                .withDeduplicationId(deduplicationId);
+        dbos.startWorkflow(
+            (dev.dbos.transact.execution.ThrowingRunnable<RuntimeException>)
```
devhawk: nit: don't use fully qualified name

transact/src/main/java/dev/dbos/transact/workflow/Debouncer.java
```diff
@@ -0,0 +1,270 @@
...
+                                + " acked but did not publish "
+                                + Constants.DEBOUNCER_CHILD_ID_KEY));
+        return dbos.retrieveWorkflow(childId);
+      }
+    }
+  }
+
+  private @Nullable String lookupExistingDebouncerId(String deduplicationId) {
+    return dbos.integration()
+        .findWorkflowIdByDeduplicationId(Constants.DBOS_INTERNAL_QUEUE, deduplicationId);
```
devhawk: use executor reference to call findWorkflowIdByDeduplicationId

transact/src/main/java/dev/dbos/transact/workflow/Debouncer.java
```diff
@@ -0,0 +1,270 @@
...
+            startOpts);
+        // Successfully enqueued a fresh debouncer for this key.
+        return dbos.retrieveWorkflow(userWorkflowId);
+      } catch (DBOSQueueDuplicatedException dup) {
+        // A debouncer for this key is already running. Forward the latest args to it.
+        // When called from inside a workflow, record the result as a durable step so that
+        // replay returns the same debouncer id and the subsequent send/getEvent steps stay
+        // deterministic. Mirrors Python's call_function_as_step("DBOS.get_deduplicated_workflow").
+        String existingDebouncerId =
+            (DBOS.inWorkflow() && !DBOS.inStep())
```
devhawk: use executor.runDbosFunctionAsStep

transact/src/main/java/dev/dbos/transact/workflow/Debouncer.java
```diff
@@ -0,0 +1,270 @@
...
+              "Debouncer for dedupId {} not found after conflict; retrying", deduplicationId);
+          continue;
+        }
+        DebouncerMessage msg = new DebouncerMessage(messageId, invocation.args(), periodMs);
+        // Send only once per call: messageId is fixed, so re-sending the same id on each
+        // retry would accumulate identical messages in the debouncer's inbox (each consuming
+        // a durable step when inside a workflow). The message persists in the notifications
+        // table until the debouncer processes it, so a single send is sufficient.
+        if (!messageSent) {
+          dbos.send(existingDebouncerId, msg, Constants.DEBOUNCER_TOPIC);
```
devhawk: send has it's own dedup infrastructure. Look at the overload that takes an idempotency key. If you use messageId as the idempotency key, you don't need to track messageSent

transact/src/main/java/dev/dbos/transact/context/DBOSContext.java
```diff
@@ -23,6 +23,9 @@ public class DBOSContext {
   private final Duration timeout;
   private final Instant deadline;
   private SerializationStrategy serialization;
+  private final Integer priority;
```
devhawk: Do we have to pass these via context? These are for workflows that are invoked directly. AppVersion in particular should not be in context - it doesn't make sense to directly invoke a different version than you're currently running

devhawk (changes requested):
Great work so far. I left a bunch of comments, but most are fairly minor (use Duration/Instant instead of longs, use registerWorkflow/startRegisteredWorkflow instead of proxies for internal workflow methods). Feel free to leave responses if you have further questions on these.

Big issues are:
* Needs a version of `Debouncer` that works with `DBOSClient`. See `DebouncerClient` in DBOS Python for reference
* `Debouncer` should have a reference to `DBOSExecutor` so it can invoke `findWorkflowIdByDeduplicationId` and `captureInvocation` directly w/o having to go thru DBOSIntegration 
* I don't think passing debounce values via context is a good idea. `appVersion` in particular is an issue since that is fixed for the running application.

transact/src/main/java/dev/dbos/transact/workflow/internal/InternalWorkflows.java
easmith: fixed

transact/src/main/java/dev/dbos/transact/workflow/internal/DebouncerServiceImpl.java
easmith: fixed

transact/src/main/java/dev/dbos/transact/workflow/internal/DebouncerServiceImpl.java
easmith: any way I make it nullable now

transact/src/main/java/dev/dbos/transact/workflow/Debouncer.java
easmith: fixed

transact/src/main/java/dev/dbos/transact/workflow/internal/InternalWorkflows.java
easmith: I've tested. You are right. fixed

transact/src/main/java/dev/dbos/transact/context/DBOSContext.java
easmith: in python lib app_version is a part of debouncer context - https://github.com/dbos-inc/dbos-transact-py/blob/main/dbos/_debouncer.py#L68

Working on it

transact/src/main/java/dev/dbos/transact/DBOS.java
easmith: implemented

transact/src/main/java/dev/dbos/transact/context/DBOSContext.java
easmith: done

transact/src/main/java/dev/dbos/transact/workflow/Debouncer.java
easmith: fixed

easmith:
Hey @devhawk
Thank you for your detailed review. I fixed all, tests works.
Could you review it again?

devhawk:
> Hey @devhawk Thank you for your detailed review. I fixed all, tests works. Could you review it again?

Will do. Note, I pushed a change to apply spotless formatting. Spotless has a [git hook](https://github.com/diffplug/spotless/tree/main/plugin-gradle#git-hook) you can install.

easmith:
One test failed. I guess it's good to rerun this test https://github.com/dbos-inc/dbos-transact-java/actions/runs/26237090589/job/77213213331?pr=390

transact/src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -84,6 +88,22 @@ public DBOS(@NonNull DBOSConfig config) {
     this.integration =
         new DBOSIntegration(
             this.config, this.workflowRegistry, dbosExecutor::get, this::registerLifecycleListener);
+    // Register the built-in debouncer service workflow directly (without a proxy) so callers can
+    // use Debouncer without having to declare and wire the service themselves.
+    var debouncerImpl = new DebouncerServiceImpl(this);
```
devhawk: This seems more complex than we need. Let's change `DebouncerServiceImpl` to something like `InternalWorkflows` and register the one workflow in it manually instead of reflecting over all the methods. This way, we don't need to implement an interface or annotate with `@Workflow`

transact/src/main/java/dev/dbos/transact/workflow/Debouncer.java
```diff
@@ -0,0 +1,332 @@
...
+    Objects.requireNonNull(wfLambda, "wfLambda must not be null");
+    if (debouncePeriod.isNegative() || debouncePeriod.isZero()) {
+      throw new IllegalArgumentException("debouncePeriod must be a positive non-zero duration");
+    }
+
+    DBOSExecutor.Invocation invocation = executor.captureInvocation(wfLambda);
+
+    // Inside a workflow, ID generation is wrapped in a step so replay is deterministic.
+    DebounceIds ids;
+    if (DBOS.inWorkflow() && !DBOS.inStep()) {
```
devhawk: I forgot executor.runDbosFunctionAsStep was private. We should make that public so it can be used here and for lookupDebouncer code below. Note, that also requires prefixing step names with "DBOS."

transact/src/main/java/dev/dbos/transact/workflow/Debouncer.java
```diff
@@ -0,0 +1,332 @@
...
+        new DebouncerOptions(
+            invocation.workflowName(),
+            invocation.className(),
+            invocation.instanceName(),
+            queueName,
+            debounceTimeout,
+            appVersion,
+            priority,
+            deduplicationId);
+    // DBOSContextHolder.get() is guaranteed non-null inside a workflow context
```
devhawk: it's always non-null

easmith: I see, of course

transact/src/main/java/dev/dbos/transact/workflow/Debouncer.java
```diff
@@ -0,0 +1,332 @@
...
+        executor.startRegisteredWorkflow(
+            debouncerWorkflow, new Object[] {options, ctx, initial}, startOpts);
+        // Successfully enqueued a fresh debouncer for this key.
+        return dbos.retrieveWorkflow(userWorkflowId);
+      } catch (DBOSQueueDuplicatedException dup) {
+        // A debouncer for this key is already running. Forward the latest args to it.
+        // When called from inside a workflow, record the result as a durable step so that
+        // replay returns the same debouncer id and the subsequent send/getEvent steps stay
+        // deterministic. Mirrors Python's call_function_as_step("DBOS.get_deduplicated_workflow").
+        String existingDebouncerId =
```
devhawk, line 278: Use executor.runDbosFunctionAsStep as per earlier comment. Probably can inline lookupExistingDebouncerId code here since it won't need to be repeated

transact/src/main/java/dev/dbos/transact/internal/DBOSIntegration.java
```diff
@@ -241,7 +259,7 @@ public Optional<RegisteredWorkflow> getRegisteredWorkflow(
    * @return an Optional containing the RegisteredWorkflow if found, otherwise empty
    */
   public Optional<RegisteredWorkflow> getRegisteredWorkflow(
-      @NonNull String workflowName, @NonNull String className, @NonNull String instanceName) {
+      @NonNull String workflowName, @NonNull String className, @Nullable String instanceName) {
```
devhawk, line 243: thank you for fixing my bug here (and related code)!

easmith: Happy to fix. If only I had the ability to do so =)

easmith:
Fixed the latest comments. I'll be mostly offline for the next few days :/

devhawk:
> Fixed the latest comments. I'll be mostly offline for the next few days :/

Cool. I'm travelling and working for a bit so my schedule is out of wack right now anyways.

transact/src/main/java/dev/dbos/transact/Constants.java
```diff
@@ -16,6 +16,14 @@ public class Constants {
 
   public static final String DBOS_INTERNAL_QUEUE = "_dbos_internal_queue";
 
+  public static final String DEBOUNCER_WORKFLOW_NAME = "_dbos_debouncer_workflow";
+  public static final String DEBOUNCER_SERVICE_CLASS_NAME =
+      "dev.dbos.transact.workflow.internal.InternalWorkflows";
```
devhawk: nitpick: Let's use `DBOS.InternalWorkflows` as the classname and `debouncerWorkflow` as the workflow name

transact/src/main/java/dev/dbos/transact/internal/DBOSIntegration.java
```diff
@@ -192,30 +194,46 @@ public Object runWorkflow(
...
   /**
-   * Get all workflows registered with DBOS.
+   * Get all user-registered workflows. Internal/system workflows registered by DBOS itself (for
+   * example, the debouncer service workflow) are excluded.
    *
-   * @return list of all registered workflow methods
+   * @return list of all user-registered workflow methods
    */
   public @NonNull Collection<RegisteredWorkflow> getRegisteredWorkflows() {
     var executor = executorSupplier.get();
```
devhawk, line 201: Since you're offline for a bit, I want to think this thru a little more. I was already thinking we might want to track internal workflows separately from user registered. At one point, the implementation of `send` was an internal workflow, but other changes eliminated the need to make `send` a workflow. Now that `debounce` is decidedly an internal workflow, we might want to track it separately. If we do that, we won't need this change to `getRegisteredWorkflows`.

transact/src/main/java/dev/dbos/transact/Constants.java
easmith: we already have DBOS_INTERNAL_QUEUE = "_dbos_internal_queue".
I would prefer existing style.

transact/src/main/java/dev/dbos/transact/internal/DBOSIntegration.java
easmith: Done. Internal workflows now live in separate maps in WorkflowRegistry/DBOSExecutor, so getRegisteredWorkflows() is back to a plain pass-through with no filtering

transact/src/main/java/dev/dbos/transact/Constants.java
devhawk: internal queue is not a namespace scoped name like workflow name + class name. We don't need "dbos" in both class and workflow name. Also, I'm not a fan of mirroring the internal class structure in the class name.

transact/src/main/java/dev/dbos/transact/workflow/Debouncer.java
devhawk: use `runDbosFunctionAsStep` instead of explicitly checking inWorkflow/inStep

transact/src/main/java/dev/dbos/transact/workflow/Debouncer.java
```diff
@@ -0,0 +1,331 @@
...
+      @NonNull ThrowingSupplier<T, E> wfLambda) {
+
+    Objects.requireNonNull(debounceKey, "debounceKey must not be null");
+    Objects.requireNonNull(debouncePeriod, "debouncePeriod must not be null");
+    Objects.requireNonNull(wfLambda, "wfLambda must not be null");
+    if (debouncePeriod.isNegative() || debouncePeriod.isZero()) {
+      throw new IllegalArgumentException("debouncePeriod must be a positive non-zero duration");
+    }
+
+    DBOSExecutor.Invocation invocation = executor.captureInvocation(wfLambda);
```
devhawk, line 232: verify the invocation WF exists via getRegisteredWorkflow

transact/src/main/java/dev/dbos/transact/workflow/internal/InternalWorkflows.java
```diff
@@ -0,0 +1,120 @@
...
+      dbos.setEvent(next.messageId(), next.messageId());
+    }
+
+    var workflow =
+        dbos.integration()
+            .getRegisteredWorkflow(
+                options.workflowName(), options.className(), options.instanceName())
+            .orElseThrow(
+                () ->
+                    new IllegalStateException(
```
devhawk: if we throw here, the debounced workflow never starts. Any workflow handles for the debounced workflow will hang in `getResult`. We need to write an ERROR workflow result row for ctx.userWorkflowId (which will require new sysdb + executor API)

transact/src/test/java/dev/dbos/transact/workflow/DebouncerTest.java
```diff
@@ -0,0 +1,362 @@
...
+            .orElse(null);
+    assertNotNull(userWfStatus, "user workflow 'process' not found on queue " + q.name());
+    assertEquals(Integer.valueOf(42), userWfStatus.priority());
+  }
+
+  // Verify that a second debounce call after the first window closes starts a fresh window.
+  // Regression test for: deduplication_id is cleared to NULL on completion, so the UNIQUE
+  // constraint no longer blocks a new enqueue with the same key.
+  @Test
+  public void reDebouncAfterWindowCloses() throws Exception {
```
devhawk: Misspelling 

```suggestion
  public void reDebounceAfterWindowCloses() throws Exception {
```

transact/src/test/java/dev/dbos/transact/client/DebouncerClientTest.java
```diff
@@ -0,0 +1,139 @@
...
+    var hB = d.debounce("key-B", Duration.ofMillis(400), "B");
+
+    assertNotEquals(hA.workflowId(), hB.workflowId());
+    assertEquals("result:A", hA.getResult());
+    assertEquals("result:B", hB.getResult());
+    assertEquals(2, serviceImpl.callCount.get());
+  }
+
+  @Test
+  void reDebouncAfterWindowCloses() throws Exception {
```
devhawk: misspelling
```suggestion
  void reDebounceAfterWindowCloses() throws Exception {
```

devhawk:
getting close. Still need to use `runDbosFunctionAsStep` in Debouncer. Also added new comments about how to handle missing workflow being debounced and regarding debouncer constants. We also need a few more tests:

- **Recovery/replay**: No test restarts DBOS mid-debounce and verifies that replay produces the same IDs and doesn't start a second user workflow. 
- **`withDeduplicationId` on the user workflow**: The `DebouncerOptions.deduplicationId` field is forwarded to the user workflow, but no test exercises it.
- **`DebouncerClient` missing a "re-debounce after window closes" test**: `DebouncerTest` has `reDebouncAfterWindowCloses` but `DebouncerClientTest` does not (the client's deduplication path is slightly different).

transact/src/test/java/dev/dbos/transact/workflow/DebouncerTest.java
easmith: fixed

transact/src/test/java/dev/dbos/transact/client/DebouncerClientTest.java
easmith: fixed

transact/src/main/java/dev/dbos/transact/workflow/internal/InternalWorkflows.java
easmith: good catch

transact/src/main/java/dev/dbos/transact/workflow/Debouncer.java
easmith: I got you now! =)

transact/src/main/java/dev/dbos/transact/Constants.java
easmith: ok. pretty clear. renamed

easmith:
> * **`DebouncerClient` missing a "re-debounce after window closes" test**: `DebouncerTest` has `reDebouncAfterWindowCloses` but `DebouncerClientTest` does not (the client's deduplication path is slightly different).

DebouncerClientTest.java:111 - already has that test

easmith:
Fixed the latest review comments and added the missing tests.
Ready to review again!

transact/src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -86,6 +90,16 @@ public DBOS(@NonNull DBOSConfig config) {
     this.integration =
         new DBOSIntegration(
             this.config, this.workflowRegistry, dbosExecutor::get, this::registerLifecycleListener);
+    // Register the built-in debouncer service workflow directly (without a proxy) so callers can
+    // use Debouncer without having to declare and wire the service themselves.
+    var internalWorkflows = new InternalWorkflows(this);
+    workflowRegistry.registerInternalInstance(internalWorkflows);
+    this.debouncerWorkflow =
```
devhawk, line 98: Call `workflowRegistry.registerInternalWorkflow` directly here. No need to route thru DBOSIntegration

transact/src/main/java/dev/dbos/transact/internal/DBOSIntegration.java
```diff
@@ -153,6 +152,29 @@ public RegisteredWorkflow registerWorkflow(
...
+   * #getRegisteredWorkflowInstances()}, but remain accessible to the executor for lookup, recovery,
+   * and dequeue.
+   *
+   * @param workflowName logical name of the internal workflow
+   * @param className name of the class that declares the workflow method
+   * @param target the singleton instance on which the method will be invoked
+   * @param method the workflow {@link Method}
+   * @throws IllegalStateException if called after DBOS is launched
+   */
+  public RegisteredWorkflow registerInternalWorkflow(
```
devhawk: This API should not be public. The one place it's called should call `workflowRegistry.registerInternalWorkflow` directly

transact/src/main/java/dev/dbos/transact/internal/DBOSIntegration.java
```diff
@@ -173,6 +195,23 @@ public RegisteredWorkflow registerWorkflow(
     return executor("startRegisteredWorkflow").startRegisteredWorkflow(regWorkflow, args, options);
   }
 
+  /**
+   * Record a terminal ERROR for a workflow that was never started, so handles awaiting it fail fast
+   * instead of polling forever. Used by the built-in debouncer workflow when it cannot start the
+   * user workflow it is responsible for.
+   */
+  public void recordErrorForUnstartedWorkflow(
```
devhawk: This API should not be public. The one place it's called should be invoking DBOSExecutor.recordErrorForUnstartedWorkflow directly

transact/src/main/java/dev/dbos/transact/workflow/internal/InternalWorkflows.java
```diff
@@ -0,0 +1,135 @@
...
+      // a terminal ERROR for the pre-assigned user workflow id. Otherwise any handle returned to
+      // the caller would poll getResult() forever, since the status row would never appear.
+      var notFound =
+          new DBOSWorkflowFunctionNotFoundException(ctx.userWorkflowId(), options.workflowName());
+      logger.error(
+          "Debouncer cannot find registered user workflow {} (id={}); recording ERROR",
+          options.workflowName(),
+          ctx.userWorkflowId(),
+          notFound);
+      dbos.integration()
```
devhawk: InternalWorkflows should get a DBOSExecutor supplier and invoke recordErrorForUnstartedWorkflow on the executor. It should not be part of the public API

transact/src/main/java/dev/dbos/transact/workflow/Debouncer.java
devhawk: Note, verification here is not needed since we write a missing workflow error in debouncerWorkflow

devhawk:
I think the remaining asks from me are:

* remove `DBOSIntegration.registerInternalWorkflow`. It's only called from DBOS constructor and it can call `workflowRegistry.registerInternalWorkflow` directly
* remove `DBOSIntegration.recordErrorForUnstartedWorkflow`. Instead, inject a `Supplier<DBOSExecutor> executorSupplier` into `InternalWorkflows` and invoke `DBOSExecutor.recordErrorForUnstartedWorkflow` directly.

transact/src/main/java/dev/dbos/transact/DBOS.java
easmith: Done

transact/src/main/java/dev/dbos/transact/workflow/internal/InternalWorkflows.java
easmith: Done

transact/src/main/java/dev/dbos/transact/workflow/Debouncer.java
easmith: removed

easmith:
Thanks for HIGH quality review. Ready to go =)

devhawk:
> Thanks for HIGH quality review. Ready to go =)

I made a small change to `debounceWorkflow` - pulled the `executorSupplier.get` call outside the while loop and checked it for null. It was minor enough that I just made the change myself.

I'm signed off but I want @kraftp to also take a look since he wrote the original debouncer code.

devhawk (approved):

transact/src/main/java/dev/dbos/transact/workflow/internal/DebouncerMessage.java
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Message sent from a {@code Debouncer} caller to the debouncer service workflow each time the
+ * debounce key fires. The debouncer service workflow uses the most recently received message's args
+ * when it eventually starts the user workflow.
+ *
+ * <p>Not part of the public API — the debouncer infrastructure consumes this directly.
+ */
+public record DebouncerMessage(
+    @NonNull String messageId, @NonNull Object[] args, @NonNull Duration debouncePeriod) {}
```
kraftp, line 15: How does typing work here? As far as I can tell, all tests have primitive types, what happens if a more complex type (user-defined object?) is routed through this?

kraftp (approved):
This looks good to me, other than one comment about typing

transact/src/main/java/dev/dbos/transact/workflow/internal/DebouncerMessage.java
devhawk: `DebouncerMessage` is used as a parameter for `debouncerWorkflow` as well as the message sent to `debouncerWorkflow`. Both of those code paths use the standard `DBOSSerializer` which is tested in `JavaSerializerTest`.

PR #391, opened 2026-05-22
Dyanmic Queue Support

### LINKED ISSUE #372: Dynamic Queues
Java version of https://github.com/dbos-inc/dbos-transact-py/pull/654

devhawk:
# PR Summary: Dynamic Queues (`devhawk/dynamic-queues`)

## New public API (`DBOS` and `DBOSClient`)
- `registerQueue(name, QueueOptions)` — creates or updates a queue persisted in the database
- `registerQueue(name, QueueOptions, QueueConflictResolution)` — same with explicit conflict handling
- `updateQueue(name, QueueOptions)` — partial update; only present fields are written, absent fields left unchanged
- `findQueue(name)` — look up a queue by name
- `listQueues()` — list all database-backed queues
- `deleteQueue(name)` — remove a queue from the database

## New types
- **`QueueOptions`** — immutable record with a `Field<T>` for each queue property, supporting three states: absent (don't touch), present-with-value, or present-with-null (clear). Fluent builder API via `set*()` static factories and `and*()` chainers.
- **`QueueConflictResolution`** — enum controlling `registerQueue` behavior: `ALWAYS_UPDATE`, `NEVER_UPDATE`, or `UPDATE_IF_LATEST_VERSION` (the last is only available on `DBOS`, not `DBOSClient`, since clients have no concept of app version).
- **`Field<T>`** — sealed interface (`Absent` / `Present`) for three-state optional values used by `QueueOptions`.

## `Queue` record changes
- Added `pollingInterval` field (defaults to 1 second) with validation
- `withRateLimit(int, double)` replaced by `withRateLimit(int, long, TimeUnit)` for type safety

## Runtime: dynamic queue polling (`QueueService`)
- New `pollDynamicQueues()` supervisor runs every second, reads all queues from the database, and starts/stops `QueueListenerTask` instances accordingly
- Static queues take precedence: if a DB queue shares a name with a static queue, the static config wins and a warning is logged
- `listenQueues` filter respected for dynamic queues too

## `SchedulerService` refactor
- `pollWorkflowSchedulesImpl` inlined; scheduler loop simplified

## Conductor protocol
- New `get_queue` / `list_queues` request+response message types and `QueueOutput` DTO, so the conductor can serve queue info to clients

## Schema / DAO
- `QueuesDAO` — full implementation: `upsertQueue`, `updateQueue`, `findQueue`, `listQueues`, `deleteQueue`, `getQueuePartitions`
- Migration updated to use `INT4` (explicitly 32-bit) for `concurrency`, `worker_concurrency`, and `rate_limit_max` columns — fixes a CRDB incompatibility where `INTEGER` defaults to 64-bit `int8`

fixes #372


transact/src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -164,6 +166,76 @@ public void registerQueues(@NonNull Queue... queues) {
...
+   * Register a database-backed dynamic queue. Must be called after launch. Queue configuration can
+   * be updated at runtime via {@link #updateQueue(String, QueueOptions)}.
+   *
+   * <p>Uses {@link QueueConflictResolution#UPDATE_IF_LATEST_VERSION} by default: the existing
+   * configuration is overwritten only if this executor is running the latest application version.
+   *
+   * @param name Queue name
+   * @param options Initial configuration options
+   */
+  public void registerQueue(@NonNull String name, @NonNull QueueOptions options) {
```
kraftp, line 179: This overload-based interface makes sense to me. In the long-term, we can deprecate the old in-memory queues.

transact/src/test/java/dev/dbos/transact/queue/DynamicQueuesTest.java
```diff
@@ -0,0 +1,273 @@
+package dev.dbos.transact.queue;
+
```
kraftp, line 2: The one thing we can add are tests that queue updates "take". The delete test needs to wait for some time for the queue thread to stop before recreating the queue, else we're not testing recreation and the test may be flaky. We also need a test for queue updates that verifies that updates to a queue's settings are picked up by its runner.

transact/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
```diff
@@ -1424,7 +1458,8 @@ private void validateWorkflow(String workflowName, String className, String inst
 
   private void validateQueue(String queueName) {
```
kraftp, line 1461: In Python and TS, we ended up essentially removing this for database-backed queues to remove the additional database read on the enqueue path

transact/src/test/java/dev/dbos/transact/queue/QueuesTest.java
```diff
@@ -336,7 +336,8 @@ public void multipleQueues() throws Exception {
   public void testLimiter() throws Exception {
 
     int limit = 5;
-    double period = 1.8; //
+    double periodSec = 1.8;
```
kraftp: In Python and TS, we converted this entire test to database-backed queues (leaving one dedicated in-memory test) to make sure there were no regressions. We may want to do the same here, as we'll eventually deprecated in-memory queues.

transact/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
devhawk: I updated queue service to keep a collection of dynamic queues as part of the polling process so we can keep this validation w/o the DB read

transact/src/test/java/dev/dbos/transact/queue/QueuesTest.java
devhawk: I renamed the original QueuesTest -> StaticQueuesTest and changed all these tests to dynamic

transact/src/test/java/dev/dbos/transact/queue/QueuesTest.java
```diff
@@ -51,6 +52,8 @@ void beforeEach() {
     dataSource = pgContainer.dataSource();
   }
 
+  // One dedicated in-memory queue test to guard against regressions on the
+  // static-queue path. All other tests below use database-backed queues.
```
kraftp: Is there a test for `listenQueue` with dynamic queues?

kraftp (approved):
Looks good, modulo one last test we need to add

transact/src/test/java/dev/dbos/transact/queue/QueuesTest.java
devhawk: DynamicQueuesTest.testListenQueue

devhawk:
> Looks good, modulo one last test we need to add

The test existed, but dynamic queue tests were spread across two files. I combined them all into the DynamicQueuesTest class. The test in question was testListenQueue

PR #400, opened 2026-05-26
@TransactionalStep

### LINKED ISSUE #385: @TransactionalStep for SpringBoot
PR #377 added support for imperative step factories, but we also need spring boot transaction integration

devhawk:
Adds `transact-spring-txstep-starter`, a new Spring Boot auto-configuration module that brings `@TransactionalStep` to Spring applications.

**What it does:**
- Introduces `@TransactionalStep`, a method annotation that marks a Spring-managed method as an idempotent DBOS step. The method body runs inside a `REQUIRES_NEW` Spring transaction; the return value is written to `tx_step_outputs` **atomically** with the user's database work. On workflow retry, the recorded output is replayed without re-executing the method body.
- `TransactionalStepAspect` intercepts annotated methods and delegates to `TransactionalStepFactory`, which calls `DBOS.runStep()` and uses `DataSourceUtils.getConnection()` to write step output on the same transaction-bound connection.
- `TransactionalStepAutoConfiguration` wires up the factory and aspect as Spring beans, auto-configured via `AutoConfiguration.imports`.
- `TransactionalStepRegistrar` scans the Spring context post-startup and calls `factory.initialize()` (creates `tx_step_outputs`) only when annotated methods are found — no DB contact for apps that don't use the annotation.
- JPA support: `TransactionalStepFactory` auto-detects `JpaTransactionManager` and sets its `dataSource` property to bridge JPA transactions to `DataSourceUtils`.

**Supported stacks (all covered by integration tests):**
- Spring JDBC / `JdbcTemplate`
- JDBI (`jdbi3-spring` transaction binding)
- jOOQ (`spring-boot-starter-jooq`)
- JPA / Hibernate (`spring-boot-starter-data-jpa`)

This PR also fixes previously implemented step factory's handling of concurrent updates. If two executors race to run the same step, both may pass the initial `checkExecution` read before either has committed. Previously, both executors transactions would commit, with the 2nd `recordOutput` INSERT being ignored. This has been fixed so the race loser's `recordOutput` INSERT into `tx_step_outputs` hits the `(workflow_id, step_id)` unique constraint. The factory catches the resulting `23505` violation and converts it to a `StepConflictException`. The calling code catches `StepConflictException`, rolls back the REQUIRES_NEW transaction (discarding the loser's database work), then re-reads tx_step_outputs to return the winner's committed result:
The outcome is idempotent from the caller's perspective: both executors return the same winner result, and only the winner's database write is committed.

Also includes minor refactors to `JdbcStepFactory`, `JdbiStepFactory`, and `JooqStepFactory` to extract shared SQL into `TxStepSchema` helpers, and expands their test suites.

fixes #385


transact-spring-txstep-starter/src/main/java/dev/dbos/transact/spring/txstep/TransactionalStepFactory.java
```diff
@@ -0,0 +1,179 @@
...
+          }
+
+          var txDef = new DefaultTransactionDefinition();
+          txDef.setPropagationBehavior(PROPAGATION_REQUIRES_NEW);
+          TransactionStatus status = txManager.getTransaction(txDef);
+          try {
+            Object result = supplier.execute();
+            Connection conn = DataSourceUtils.getConnection(dataSource);
+            recordOutput(conn, workflowId, stepId, result);
+            txManager.commit(status);
```
kraftp, line 179: One thing Claude flagged--not sure if true, worth verifying--is that in JPA it's common for the commit itself to throw (for example if there's a primary key error on an insert) and if that happens, the rollbacks will throw and break this logic. Might add a test for that in JPA.

transact-spring-txstep-starter/src/main/java/dev/dbos/transact/spring/txstep/TransactionalStepAutoConfiguration.java
```diff
@@ -0,0 +1,71 @@
...
+   * Applies the JPA bridge when {@code spring-orm} is on the classpath. If a {@code
+   * JpaTransactionManager} is in use and its {@code dataSource} property is not set, sets it so
+   * that {@code DataSourceUtils.getConnection(dataSource)} returns the transaction-bound connection
+   * inside {@code @TransactionalStep} methods.
+   */
+  @Configuration(proxyBeanMethods = false)
+  @ConditionalOnClass(name = "org.springframework.orm.jpa.JpaTransactionManager")
+  static class JpaBridgeConfiguration {
+    @Bean
+    JpaBridgeApplier jpaBridgeApplier(PlatformTransactionManager txManager, DataSource dataSource) {
```
kraftp: Is there a better way to do this than an empty bean?

transact-spring-txstep-starter/src/main/java/dev/dbos/transact/spring/txstep/TransactionalStepAutoConfiguration.java
```diff
@@ -0,0 +1,71 @@
...
+ * <p>Activates when a {@link DBOS} bean, a {@link PlatformTransactionManager} bean, and a {@link
+ * DataSource} bean are all present. Creates {@link TransactionalStepFactory}, {@link
+ * TransactionalStepAspect}, and {@link TransactionalStepRegistrar} beans.
+ *
+ * <p>If {@code JpaTransactionManager} is detected and its {@code dataSource} property is not yet
+ * set, it is set automatically so that {@code DataSourceUtils.getConnection()} returns the
+ * transaction-bound connection inside {@code @TransactionalStep} methods.
+ */
+@AutoConfiguration(after = DBOSAutoConfiguration.class)
+@ConditionalOnBean({DBOS.class, PlatformTransactionManager.class, DataSource.class})
```
kraftp, line 35: Claude flags this as fragile because it has no ordering declared relative to `DataSource` or `PlatformTransactionManager`. Again, not sure how real this is, but worth checking.

kraftp (commented):
Looks good overall, the new interface is great for Spring! I did an AI review and it flagged some idiomatic Java things, probably worth checking them but they could be nothing.

transact-spring-txstep-starter/README.md
```diff
@@ -0,0 +1,192 @@
...
+`@TransactionalStep` marks a Spring-managed method as an **idempotent transactional step** inside a DBOS workflow. Each call runs in a `REQUIRES_NEW` Spring transaction; the step's return value is written to a `tx_step_outputs` table **atomically** with the user's database work. On workflow retry, the recorded output is replayed without re-executing the method body.
+
+## Installation
+
+Add both the DBOS Spring Boot starter and this module:
+
+**Gradle**
+```kotlin
+implementation("dev.dbos:transact-spring-boot-starter:<version>")
+implementation("dev.dbos:transact-spring-txstep-starter:<version>")
```
kraftp, line 14: Why is this thing called "starter" if it's the actual, production-ready (presumably) implementation?

transact-spring-txstep-starter/src/main/java/dev/dbos/transact/spring/txstep/TransactionalStepFactory.java
```diff
@@ -0,0 +1,179 @@
...
+   *
+   * <p>The step output is written to {@code tx_step_outputs} atomically with the transaction
+   * started for the user's work. On failure the transaction is rolled back and the error is
+   * recorded in a separate connection so retries can replay the error without re-executing.
+   *
+   * @param supplier the step body; typically {@code () -> pjp.proceed()} from the aspect
+   * @param stepName stable name for the step within the workflow
+   */
+  @SuppressWarnings("unchecked")
+  public <E extends Exception> Object runTransactionalStep(
```
kraftp, line 141: What does this do if called from within an existing Spring transaction?

transact-spring-txstep-starter/src/main/java/dev/dbos/transact/spring/txstep/TransactionalStepAutoConfiguration.java
devhawk: it's real. Found it independently when I was trying to update widget store

transact-spring-txstep-starter/README.md
devhawk: that's spring boot naming convention. For example, the Spring Web dependency is `spring-boot-starter-webmvc`

transact-spring-txstep-starter/src/main/java/dev/dbos/transact/spring/txstep/TransactionalStepFactory.java
devhawk: @TransactionalStep requires a new transaction, but it's worth adding tests for

devhawk: Are you asking what happens if a `@TransactionalStep` is called from outside a workflow? We should also add tests for that

transact-spring-txstep-starter/src/main/java/dev/dbos/transact/spring/txstep/TransactionalStepAutoConfiguration.java
devhawk: AFAIK, no. JPA (aka Hibernate) builds on top of the vanilla `PlatformTransactionManager` so we need the extra bridge code to ensure JPA code is executed in the transaction we specify. I will double check on this tomorrow

transact-spring-txstep-starter/src/main/java/dev/dbos/transact/spring/txstep/TransactionalStepFactory.java
kraftp: Yeah, let's test that case

transact-spring-txstep-starter/src/main/java/dev/dbos/transact/spring/txstep/TransactionalStepFactory.java
devhawk: I will add a test

transact-spring-txstep-starter/src/main/java/dev/dbos/transact/spring/txstep/TransactionalStepAutoConfiguration.java
devhawk: It turns out JPA handles this automatically. Unneeded code removed

transact-spring-txstep-starter/src/main/java/dev/dbos/transact/spring/txstep/TransactionalStepFactory.java
devhawk: I've updated the code to so that @TransactionalStep acts like @Transactional if it's executed outside of a workflow or inside of a step. Note, that means that if you call a @TransactionalStep from a @TransactionalStep, they both get executed in the same transaction (i.e. the tx created for the outer tx step automatically flows into the nested tx step) but only the outer tx step gets durably executed. 

Also, note that the default behavior of @Transactional is to rollback on runtime exceptions and commit on checked exceptions. We mirror that behavior in the `!DBOS.inWorkflow() || DBOS.inStep()` branch.

added outsideWorkflow* and nestedStep* tests

transact-spring-txstep-starter/src/main/java/dev/dbos/transact/spring/txstep/TransactionalStepFactory.java
devhawk: verified. adding !status.isCompleted checks before calling txManager.commit / rollback to avoid + adding commitTimeException_rollsBackAndRecordsError test

transact-spring-txstep-starter/README.md
```diff
@@ -0,0 +1,192 @@
...
+        return order;
+    }
+}
+```
+
+`spring-boot-starter-jooq` wraps the `DataSource` in a `TransactionAwareDataSourceProxy`, so jOOQ queries execute on the Spring-transaction-bound connection.
+
+### JPA / Hibernate
+
+Spring Boot auto-configures `JpaTransactionManager` when `spring-boot-starter-data-jpa` is present. `@TransactionalStep` detects `JpaTransactionManager` and sets its `dataSource` property automatically if not already configured, bridging JPA transactions to `DataSourceUtils`.
```
kraftp: Is this still true?

kraftp (approved):
Looks good!

transact-spring-txstep-starter/README.md
devhawk: No, it happens automatically without us. Will fix

PR #403, opened 2026-05-27
SendBulk

### LINKED ISSUE #402: Bulk Send
Java version of https://github.com/dbos-inc/dbos-transact-py/pull/690

devhawk:
Adds `sendBulk()` to the public API (DBOS and DBOSClient), allowing a list of messages to be sent to one or more workflow destinations atomically in a single database transaction. Also adds `sendToForks` support to both `send()` and `sendBulk()`, which fans out each message to every workflow recursively forked from the destination.

fixes #402

### DIFF
--- a/transact/src/main/java/dev/dbos/transact/DBOS.java
+++ b/transact/src/main/java/dev/dbos/transact/DBOS.java
@@ -17,6 +17,7 @@ import dev.dbos.transact.workflow.Queue;
 import dev.dbos.transact.workflow.QueueConflictResolution;
 import dev.dbos.transact.workflow.QueueOptions;
 import dev.dbos.transact.workflow.ScheduleStatus;
+import dev.dbos.transact.workflow.SendMessage;
 import dev.dbos.transact.workflow.SerializationStrategy;
 import dev.dbos.transact.workflow.Step;
 import dev.dbos.transact.workflow.StepInfo;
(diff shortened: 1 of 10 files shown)


transact/src/main/java/dev/dbos/transact/database/dao/NotificationsDAO.java
```diff
@@ -35,48 +45,136 @@ private NotificationsDAO() {}
...
+          destinations.add(baseDest);
+          if (sendToForks) {
+            var desc = forkDescendants.getOrDefault(baseDest, Set.of());
+            desc.stream().sorted().forEach(destinations::add);
+          }
+
+          for (String dest : destinations) {
+            String uuid;
+            if (msg.idempotencyKey() != null) {
+              uuid = sendToForks ? msg.idempotencyKey() + "::" + dest : msg.idempotencyKey();
```
kraftp: We ended up changing this to always suffix. That ways idempotency keys are per-workflow instead of global and there isn't a gotcha where they're different depending on what options you're using.

transact/src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -541,8 +542,35 @@ public void send(
       @Nullable String topic,
       @Nullable String idempotencyKey,
       @Nullable SerializationStrategy serialization) {
+    send(destinationId, message, topic, idempotencyKey, serialization, false);
+  }
+
+  public void send(
```
kraftp, line 559: No Javadocs, unlike everything else

kraftp (approved):

PR #410, opened 2026-06-02
ExternalState.updateTime Instant

devhawk:
** BREAKING CHANGE ** (limited scope to event receivers)

Changes ExternalState.updateTime from a BigDecimal to an Instant. Stores the value in the `NUMERIC(38,15)` update_time column as fractional milliseconds. This provides backwards compatibility (previously we expected users to store whole epoch ms in update_time) while also providing full Instant nanosecond precision.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,6 @@ bin/
 
 ### Mac OS ###
 .DS_Store
+
+### Scala Metals LSP files
+.metals/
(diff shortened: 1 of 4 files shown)


kraftp (approved):

devhawk:
Two demo apps failed. Merging anyway

* java/dbos-starter failed to pull a docker container
* java/widget-store tests are broken. Will fix in demo app repo

PR #424, opened 2026-06-05
Eliminate ASM dependency

### LINKED ISSUE #422: Revist ASM dependency

devhawk:
ASM is a powerful byte code manipulation library. We were only using it to parse class files when hashing registered workflow methods for a generated an app version. This PR manually writes that class parsing file so we can drop the ASM dependency 

fixes #422

### DIFF
--- a/transact/src/main/java/dev/dbos/transact/internal/AppVersionComputer.java
+++ b/transact/src/main/java/dev/dbos/transact/internal/AppVersionComputer.java
@@ -24,76 +24,84 @@ public class AppVersionComputer {
       final var hasher = MessageDigest.getInstance("SHA-256");
       hasher.update(dbosVersion.getBytes(StandardCharsets.UTF_8));
 
-      var sortedWorkflows =
+      var workflowIterator =
           workflows.stream()
               .sorted(Comparator.comparing(RegisteredWorkflow::fullyQualifiedName))
               .iterator();
 
-      while (sortedWorkflows.hasNext()) {
-        var wf = sortedWorkflows.next();
+      while (workflowIterator.hasNext()) {
+        var wf = workflowIterator.next();
         var klass = wf.workflowMethod().getDeclaringClass();
         var klassPath = klass.getName().replace('.', '/') + ".class";
         var methodDesc = getMethodDescriptor(wf.workflowMethod());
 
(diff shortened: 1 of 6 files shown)


transact/src/main/java/dev/dbos/transact/internal/AppVersionComputer.java
```diff
@@ -28,176 +24,215 @@ public static String computeAppVersion(
...
-
-    public HashingMethodVisitor(MessageDigest md) {
-      super(Opcodes.ASM9);
-      this.md = md;
-    }
-
-    private int labelOrdinal(Label label) {
-      return labelOrdinals.computeIfAbsent(label, l -> nextLabelOrdinal++);
-    }
+  /**
```
kraftp, line 55: How much of a problem is this to maintain? Like, will this break when they release a new Java version?

devhawk: The class file format for the purposes of hashing the method byte codes has been stable since Java 11. There are upcoming changes (Project Vahalla) but there is also a new [class file API](https://openjdk.org/jeps/484). That API was shipped in JDK 24, so it's not available to us (we target JDK 17). 

There is a [backport](https://github.com/smallrye/jdk-classfile-backport) available. using it would mean either swapping the ASM dependency for jdk-classfile-backport. Not sure which would be harder to explain. ASM is much more powerful but I suspect jdk-classfile-backport is much less well known.

kraftp: Got it! So the approach we're currently taking makes sense then, as long as we're confident in its stability. We should also make sure that if it does fail, the failure is localized, and it doesn't run at all if you pass in a custom app version.

devhawk: we return `"unknown-" + System.currentTimeMillis();` if there's any exception. This code path is only called if appVersion is null or empty at DBOS Launch. `DBOS__APPVERSION` env var, `DBOSConfig.appVersion` and `DBOSConfig.enablePatching` would all have the be null/empty (or false in the case of `enablePatching`) for appVersion to be null at DBOS launch

kraftp (approved):

PR #425, opened 2026-06-08
fork from failure

### LINKED ISSUE #351: Fork From Failure support

devhawk:
> Note, this PR has a minor breaking change detailed below

This PR implements forkFromFailure, which re-executes a failed workflow starting from a specific prior step, copying completed step outputs so they don't re-run. 

forkWorkflow and forkFromFailure now share a bulkForkWorkflows helper which owns transaction management and handles batch inserts

ForkOptions timeout field changed from Timeout to @Nullable Duration — fork timeouts are always explicit or absent, never inherited. ForkOptions.timeout is a public field used by the API so this is a minor breaking change

fixes #351

### DIFF
--- a/transact/src/main/java/dev/dbos/transact/DBOS.java
+++ b/transact/src/main/java/dev/dbos/transact/DBOS.java
@@ -13,7 +13,6 @@ import dev.dbos.transact.internal.QueueRegistry;
 import dev.dbos.transact.internal.WorkflowRegistry;
 import dev.dbos.transact.migrations.MigrationManager;
 import dev.dbos.transact.workflow.Debouncer;
-import dev.dbos.transact.workflow.ForkFromFailureOptions;
 import dev.dbos.transact.workflow.ForkOptions;
 import dev.dbos.transact.workflow.ListWorkflowsInput;
 import dev.dbos.transact.workflow.Queue;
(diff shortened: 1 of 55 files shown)


kraftp (approved):
Looks good--does this pass Conductor tests?

transact/src/main/java/dev/dbos/transact/DBOS.java
```diff
@@ -941,6 +942,35 @@ public void deleteWorkflows(@NonNull List<String> workflowIds, boolean deleteChi
...
+  /**
+   * Fork a single workflow from the point determined by the given options. The mode
+   * (fromLastFailure, fromLastStep, fromStep, or fromStepName) controls which step the fork
+   * re-executes from.
+   *
+   * @param workflowId ID of the workflow to fork
+   * @param options Fork-from-failure options; exactly one mode must be set
+   * @return handle to the newly forked workflow
+   */
+  public @NonNull WorkflowHandle<Object, Exception> forkFromFailure(
```
kraftp: If we're going to make it public, call it something like `bulkFork`. Might go back and change the internal Python/TS name to that (not exposed publicly in those languages).

devhawk:
(closed by accident, reopened)

transact/src/main/java/dev/dbos/transact/DBOS.java
devhawk: I'll remove the public API

PR #427, opened 2026-06-09
Step Should Retry support

### LINKED ISSUE #413: Configurable Error Retrying For Steps
Analogous to https://docs.dbos.dev/python/tutorials/step-tutorial#filtering-retries-with-should_retry

devhawk:
Adds a shouldRetry predicate to steps, allowing callers to short-circuit retries on specific exception types rather than always exhausting maxAttempts.

Two ways to specify the predicate:
- Annotation (`@Step`): provide a `Class<? extends StepShouldRetry>` — the class is instantiated via reflection at runtime. StepShouldRetry is a functional interface w/ `boolean shouldRetry(Throwable)` method.

```java
@Step(maxAttempts = 5, shouldRetry = MyRetryPolicy.class)
public String myStep(String input) { ... }
```

- Programmatic (`StepOptions`): pass a Predicate<Throwable> directly, suitable for inline steps or lambdas.

```java
dbos.runStep(
  () -> { ... },
  new StepOptions("myStep")
    .withMaxAttempts(5)
    .withShouldRetry(e -> !(e instanceof FatalException)));
```
  When shouldRetry returns false, the step aborts immediately and rethrows the original exception. When it returns true (or is not set), normal retry-with-backoff behavior applies.

Implementation notes
  - StepShouldRetry.None is a sentinel inner class used as the annotation default, since Java annotations cannot use null as an element default value.
  - StepShouldRetry implementers must have a public no-arg constructor

fixes #413

### DIFF
--- a/transact/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
+++ b/transact/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
@@ -1210,8 +1210,18 @@ public class DBOSExecutor implements AutoCloseable {
       }
 
       var shouldRetry = options.shouldRetry();
-      if (exception != null && shouldRetry != null && !shouldRetry.test(exception)) {
-        break;
+      if (exception != null && shouldRetry != null) {
+        try {
+          if (!shouldRetry.test(exception)) {
+            break;
+          }
+        } catch (Exception predicateEx) {
+          logger.warn(
+              "shouldRetry predicate threw for step {} in workflow {}; treating as retry",
+              options.name(),
+              workflowId,
+              predicateEx);
+        }
       }
 
       try {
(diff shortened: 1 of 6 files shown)


transact/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
```diff
@@ -1209,6 +1209,11 @@ private <T, E extends Exception> T runStepInternal(
         break;
       }
 
+      var shouldRetry = options.shouldRetry();
```
kraftp, line 1212: What happens if the predicate itself fails/throws an exception for whatever reason?

devhawk: updating this code to catch exceptions thrown by `shouldRetry.test`, log them and continue retrying (i.e. we treat an exception thrown by shouldRetry.test as returning `true`)

kraftp (approved):

PR #428, opened 2026-06-10
Listen/Notify for readStream

### LINKED ISSUE #401: Listen/Notify in Streams
Java version of https://github.com/dbos-inc/dbos-transact-py/pull/691

devhawk:
Replaces the 1-second Thread.sleep polling loop in StreamIterator with PostgreSQL LISTEN/NOTIFY, so stream readers wake immediately when a new value is inserted. The 1-second poll is retained as a fallback timeout and for use when LISTEN/NOTIFY is not available (like CRDB)

A previous PR added Migration 39 which adds the trigger that fires `pg_notify('dbos_streams_channel', workflow_uuid || '::' || key)` on every INSERT to the streams table. This PR wires up the reader side.

fixes #401

### DIFF
--- a/transact/src/main/java/dev/dbos/transact/database/dao/StreamsDAO.java
+++ b/transact/src/main/java/dev/dbos/transact/database/dao/StreamsDAO.java
@@ -156,49 +156,29 @@ public class StreamsDAO {
 
     while (true) {
       ctx.checkClosed();
-      try (var sub = notificationRegistry.subscribe(new SignalKey.Stream(workflowId, key));
-          var conn = ctx.getConnection();
-          var stmt = conn.prepareStatement(sql)) {
-        stmt.setString(1, workflowId);
-        stmt.setString(2, key);
-        stmt.setInt(3, offset);
-        try (var rs = stmt.executeQuery()) {
-          if (rs.next()) {
-            String value = rs.getString("value");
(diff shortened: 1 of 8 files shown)


transact/src/main/java/dev/dbos/transact/database/dao/StreamsDAO.java
```diff
@@ -142,23 +154,31 @@ public static Object readStream(DbContext ctx, String workflowId, String key, in
...
-      stmt.setString(2, key);
-      stmt.setInt(3, offset);
-      try (var rs = stmt.executeQuery()) {
-        if (rs.next()) {
-          String value = rs.getString("value");
-          String serialization = rs.getString("serialization");
-          Object deserialized = SerializationUtil.deserializeValue(value, serialization, null);
-          if (STREAM_CLOSED_SENTINEL.equals(deserialized)) {
-            throw new IllegalStateException("Stream closed for key: " + key);
+    try (var sub = notificationRegistry.subscribe(new SignalKey.Stream(workflowId, key))) {
```
kraftp: This isn't reset in the loop. So if there's any kind of stale notification on the key (which can happen, you read N directly before getting the signal for N), this will spin until N+1 arrives.

kraftp (approved):

PR #430, opened 2026-06-11
Cancellation semantics + signal map recfactoring

### LINKED ISSUE #414: Improve Cancellation Semantics
Allow cancelling children (https://github.com/dbos-inc/dbos-transact-py/pull/701) and make cancellation a terminal state (https://github.com/dbos-inc/dbos-transact-py/pull/703)

devhawk:
Added cancellation checks in recv and getEvent .

Alligned cancelation semantics to TS/Python. WorkflowCancelled is thrown inside the running workflow to abort it. AwaitedWorkflowCancelled is thrown to external callers requesting a result.

Simplied SignalMap, removed generic argument in favor of String and removed WakeReason.

fixes #414


transact/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
```diff
@@ -1752,8 +1755,13 @@ private <T, E extends Exception> WorkflowHandle<T, E> executeWorkflow(
 
             logger.error("executeWorkflow {}", workflowId, actual);
 
-            if (actual instanceof InterruptedException
-                || actual instanceof DBOSWorkflowCancelledException) {
+            if (actual instanceof DBOSWorkflowCancelledException cancelled) {
+              if (cancelled.workflowId().equals(workflowId)) {
+                throw cancelled;
+              } else {
+                throw new DBOSAwaitedWorkflowCancelledException(cancelled.workflowId());
```
kraftp: But then this isn't recorded, which leaves the workflow PENDING forever

transact/src/test/java/dev/dbos/transact/notifications/NotificationServiceTest.java
```diff
@@ -531,6 +531,30 @@ public void sendFromStep() throws Exception {
...
+    dbos.cancelWorkflow(handle.workflowId());
+
+    WorkflowState finalState = null;
+    long deadline = System.currentTimeMillis() + 10_000;
+    while (System.currentTimeMillis() < deadline) {
+      finalState = dbos.retrieveWorkflow(handle.workflowId()).getStatus().status();
+      if (finalState == WorkflowState.CANCELLED) break;
+      Thread.sleep(100);
+    }
+    assertEquals(WorkflowState.CANCELLED, finalState);
```
kraftp: Of course it starts cancelled, but does is end that way? Does the workflow actually terminate?

transact/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
devhawk: This doesn't change the semantics of recording. It only changes which exception is thrown.

kraftp: But if an error is thrown here the recording code doesn't run, right? If I'm reading that function correctly. Let's add a test for it, testing what happens if DBOSWorkflowCancelledException from another ID is thrown in a workflow.

transact/src/test/java/dev/dbos/transact/notifications/NotificationServiceTest.java
```diff
@@ -531,6 +532,25 @@ public void sendFromStep() throws Exception {
...
+
+    // Wait for recv to signal it's ready, then give it a moment to enter the polling loop
+    impl.recvReadyLatch.await();
+    Thread.sleep(200);
+
+    dbos.cancelWorkflow(handle.workflowId());
+
+    // Blocks until the workflow task actually exits, proving the recv loop detected the
+    // cancellation and the workflow thread terminated — not just that the DB was updated.
+    assertThrows(DBOSAwaitedWorkflowCancelledException.class, handle::getResult);
```
kraftp, line 551: This should also assert the workflow completes (`getResult` on the original handle) in reasonable time.

kraftp (approved):

transact/src/test/java/dev/dbos/transact/notifications/NotificationServiceTest.java
devhawk: This line is calling getResult on the original handle to assert that it throws. All the tests have a [default 2 minute timeout](https://github.com/dbos-inc/dbos-transact-java/blob/main/transact/src/test/resources/junit-platform.properties#L7). 

Do I need more than that?

kraftp: Got it, looks good!

PR #431, opened 2026-06-12
Auth Context

### LINKED ISSUE #167: Auth Support
authenticatedUser / authenticatedRoles / assumedRole

devhawk:
Adds authentication context (authenticatedUser, assumedRole, authenticatedRoles) as a first-class concept throughout the workflow execution stack, so that identity established at the entry point is automatically carried through to child workflows, queued execution, and recovery.

New APIs:
- DBOS.authenticatedUser(), DBOS.assumedRole(), DBOS.authenticatedRoles() — static accessors to read auth context from within a running workflow
- WorkflowOptions.withAuthentication(user, roles) / .withAuthenticatedUser() / .withAssumedRole() / .withAuthenticatedRoles() — set auth when starting a workflow inline
- DBOSClient.EnqueueOptions.withAuthentication(user, roles) / .withAuthenticatedUser() / .withAssumedRole() / .withAuthenticatedRoles() — set auth when enqueueing remotely
- StartWorkflowOptions gains the same three auth fields

Propagation
- Auth fields stored in DBOSContext and carried forward when spawning child workflows (both direct invocation and startWorkflow)
- Auth fields written to the workflow status row at persist time and restored on recovery, so identity is durable across crashes
- Enqueued workflows carry auth through to dequeue and execution


transact/src/main/java/dev/dbos/transact/context/DBOSContext.java
```diff
@@ -140,32 +158,70 @@ public SerializationStrategy getSerialization() {
...
+      resolvedDeadline = null;
+    } else if (nextTimeout instanceof Timeout.Explicit e) {
+      resolvedTimeout = e.value();
+      resolvedDeadline = Instant.ofEpochMilli(System.currentTimeMillis() + e.value().toMillis());
+    }
+    return new TimeoutAndDeadline(resolvedTimeout, resolvedDeadline);
+  }
+
+  public String getAuthenticatedUser() {
+    return nextAuthenticatedUser != null ? nextAuthenticatedUser : authenticatedUser;
```
kraftp: I don't fully understand this--why `nextAuthenticatedUser` and what is that?

devhawk: it's so you can set the auth info for the workflow you're about to run from the auth info for the currently running workflow.

Your comment made me realize I didn't have a mechanism for clearing the auth info for the next workflow. So a scenario where parent workflow has auth info, but it wants to directly invoke a child workflow w/o auth info was not supported. Fixed.

transact/src/main/java/dev/dbos/transact/context/WorkflowOptions.java
```diff
@@ -84,12 +107,113 @@ public WorkflowOptions(@Nullable String workflowId) {
...
+  }
+
+  /** Create a WorkflowOptions like this one, but with the authenticated user and roles cleared */
+  public @NonNull WorkflowOptions withNoAuthentication() {
+    return new WorkflowOptions(
+        this.workflowId,
+        this.timeout,
+        this.deadline,
+        Field.of(null),
+        this.assumedRole,
```
kraftp: Shouldn't this clear the roles?

kraftp (approved):

transact/src/main/java/dev/dbos/transact/context/WorkflowOptions.java
devhawk: yeah it should also clear assumedRole

PR #437, opened 2026-06-16
revert strongly typed aggregates input status

devhawk:
Revert part of #435 that strongly typed the status field of aggregates input records


devhawk (commented):
Note, I ran the conductor tests against this branch locally, and they all passed

kraftp (approved):

PR #443, opened 2026-06-19
start workflow options auth clear override mechanism

### LINKED ISSUE #439: start workflow options auth clear override mechanism
there's no way to specify "no authenticated user" in a child workflow from a parent workflow with a child workflow via start workflow options

devhawk:
Fixes #439


devhawk:
we've decided not to do this work at this time. If users ask for this down the road, we will revisit

PR #450, opened 2026-06-30
Fix concurrent applySchedules invocation

devhawk:
Port of https://github.com/dbos-inc/dbos-transact-py/pull/741 to Java.

# Problem

SchedulesDAO.applySchedules implemented "create or replace" as DELETE followed by INSERT (with a freshly generated schedule_id every time). Two concurrent applySchedules calls for the same schedule name could interleave their delete/insert pairs and collide on the schedule_name unique constraint. The new schedule_id on every apply also meant a schedule's identity churned on every edit.

# Fix
- SchedulesDAO: replaced delete+insert with a single INSERT ... ON CONFLICT (schedule_name) DO UPDATE, making applySchedules idempotent and race-free. On conflict, schedule_id, status, and last_fired_at are preserved from the existing row; only the definition fields (workflow name/class, cron, context, backfill flag, timezone, queue) are updated.
- SchedulerService: since schedule_id no longer changes on re-apply, the poller can no longer rely on a new ID to detect an edited schedule. Added RunningSchedule, a record snapshotting each running schedule's definition fields (excluding identity/status/runtime state) plus its current ScheduledFuture, keyed by schedule id in a single map (runningSchedules). Each poll compares the live schedule against the snapshot via RunningSchedule.matches(...); on a mismatch, the schedule's future is cancelled and restarted with the new definition (no backfill on restart — only on first start).

### DIFF
--- a/transact/src/main/java/dev/dbos/transact/execution/SchedulerService.java
+++ b/transact/src/main/java/dev/dbos/transact/execution/SchedulerService.java
@@ -34,11 +34,10 @@ import org.slf4j.LoggerFactory;
 
 public class SchedulerService implements AutoCloseable {
 
-  // Tracks a running schedule: the definition fields are a snapshot used to detect definition
-  // changes, the future is the currently scheduled task which is replaced on every firing.
-  // Context is compared as its raw serialized string (from ScheduleRecord) rather than the
-  // deserialized value, since the deserialized value's class may not implement a meaningful
-  // equals().
+  // Tracks a running schedule: the definition fields are a snapshot of the schedule fields used to
+  // detect changes, the future is the currently scheduled task which is replaced on every firing.
+  // Context is compared as its raw serialized string since the deserialized value's class may not
+  // implement a meaningful equals().
   private record RunningSchedule(
       String workflowName,
       String className,
@@ -46,7 +45,10 @@ public class SchedulerService implements AutoCloseable {
       String context,
       ZoneId cronTimezone,
       String queueName,
-      AtomicReference<ScheduledFuture<?>> future) {
+      AtomicReference<ScheduledFuture<?>> future,
+      // Set by cancelWorkflowSchedule so a task that's already mid-execution won't reschedule
+      // itself after being removed from the map.
+      AtomicBoolean canceled) {
 
     RunningSchedule(ScheduleRecord record) {
       this(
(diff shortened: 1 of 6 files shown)


transact/src/main/java/dev/dbos/transact/execution/SchedulerService.java
```diff
@@ -33,6 +33,38 @@
...
+          schedule.cronTimezone(),
+          schedule.queueName(),
+          new AtomicReference<>());
+    }
+
+    boolean matches(WorkflowSchedule schedule) {
+      return Objects.equals(workflowName, schedule.workflowName())
+          && Objects.equals(className, schedule.className())
+          && Objects.equals(cron, schedule.cron())
+          && Objects.equals(context, schedule.context())
```
kraftp: Is this comparison stable in Java?

transact/src/main/java/dev/dbos/transact/execution/SchedulerService.java
```diff
@@ -302,9 +353,12 @@ private ScheduledFuture<?> scheduleTask(ZonedDateTime nextTime, Runnable task) {
   }
 
   private void cancelWorkflowSchedule(String scheduleId) {
-    var future = workflowScheduleFutures.remove(scheduleId);
-    if (future != null) {
-      future.cancel(false);
+    var running = runningSchedules.remove(scheduleId);
```
kraftp, line 375: I strongly suspect this can race with the schedule run function. Lots of checks of shared variables.

transact/src/main/java/dev/dbos/transact/execution/SchedulerService.java
devhawk: the context value is *not* comparison stable - it depends on the .equals method of the underlying type. will fix

transact/src/main/java/dev/dbos/transact/execution/SchedulerService.java
devhawk: Let's discuss @ standup

devhawk: yup, there was a race. RunningSchedule needed to track cancellation status independently of inclusion in running schedules map

kraftp (approved):

PR #452, opened 2026-07-01
No-version workflows are only dequeued by latest-version executors

maxdml:


transact/src/main/java/dev/dbos/transact/database/dao/QueuesDAO.java
```diff
@@ -134,15 +134,19 @@ SELECT COUNT(*)
...
             FROM "%s".workflow_status
             WHERE queue_name = ?
               AND status = ?
-              AND (application_version = ? OR application_version IS NULL)
+              AND (application_version = ? OR (application_version IS NULL AND ? = (
+                    SELECT version_name FROM "%s".application_versions
+                    ORDER BY version_timestamp DESC LIMIT 1)))
           """
-              .formatted(ctx.schema());
+              .formatted(ctx.schema(), ctx.schema());
```
devhawk: use "%1$s" instead of "%s" in the query string so we only have to pass the schema value once

transact/src/main/java/dev/dbos/transact/database/dao/QueuesDAO.java
```diff
@@ -134,15 +134,36 @@ SELECT COUNT(*)
         maxTasks = Math.min(maxTasks, availableTasks);
       }
 
+      // Version-less workflows (application_version IS NULL) are only dequeued
+      // when this worker is running the latest registered application version.
+      boolean isLatestVersion = true;
+      String latestVersionQuery =
+          """
+            SELECT version_name FROM "%1$s".application_versions
```
devhawk: 1$ syntax only needed for using a substitution more than once

transact/src/main/java/dev/dbos/transact/database/dao/QueuesDAO.java
```diff
@@ -134,15 +134,36 @@ SELECT COUNT(*)
...
+      String versionClause =
+          isLatestVersion
+              ? "(application_version = ? OR application_version IS NULL)"
+              : "application_version = ?";
+
       var query =
           """
             SELECT workflow_uuid
-            FROM "%s".workflow_status
+            FROM "%1$s".workflow_status
```
devhawk: 1$ syntax only needed for using a substitution more than once. We should be able to just use %s for each marker

devhawk:
Fine by me if it's fine w/ @kraftp

devhawk (approved):

PR #456, opened 2026-07-13
Active lock management

maxdml:
Release lock on active IDs before writing workflow outcome.

### DIFF
--- a/transact/src/main/java/dev/dbos/transact/database/dao/WorkflowDAO.java
+++ b/transact/src/main/java/dev/dbos/transact/database/dao/WorkflowDAO.java
@@ -336,13 +336,13 @@ public class WorkflowDAO {
           "updateWorkflowOutcome called with non-terminal status: " + status);
     }
 
-    // Never overwrite a CANCELLED workflow: a workflow cancelled during its final step must not
-    // subsequently complete.
+    // Only a PENDING row can receive an outcome: any other status means this run was
+    // superseded (cancelled during its final step, re-enqueued by a concurrent resume, ...).
     var sql =
         """
           UPDATE "%s".workflow_status
           SET status = ?, output = ?, error = ?, updated_at = ?, completed_at = ?, deduplication_id = NULL
-          WHERE workflow_uuid = ? AND status != ?
+          WHERE workflow_uuid = ? AND status = ?
         """
             .formatted(schema);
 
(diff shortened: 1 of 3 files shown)


transact/src/main/java/dev/dbos/transact/database/dao/WorkflowDAO.java
```diff
@@ -367,8 +368,12 @@ static void updateWorkflowOutcome(
         try (var readStmt = conn.prepareStatement(readSql)) {
           readStmt.setString(1, workflowId);
           try (var rs = readStmt.executeQuery()) {
-            if (rs.next() && WorkflowState.CANCELLED.name().equals(rs.getString(1))) {
-              throw new DBOSWorkflowCancelledException(workflowId);
+            if (rs.next()) {
+              var current = rs.getString(1);
+              if (!WorkflowState.SUCCESS.name().equals(current)
```
devhawk: I don't understand the logic here. If the workflow was superseded (i.e. any other state than success/error) instead of cancelled, why raise a cancelled exception?

maxdml: Today I find it easier to return a cancel error to cover all the cases:
- Workflow is cancelled
- Workflow is enqueued, which means it was resumed (presumably after a cancel) and is or will be executed on a different routine
- Workflow is in DLQ, which could throw an other exception if we really wanted to

transact/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
```diff
@@ -1746,6 +1746,17 @@ private <T, E extends Exception> WorkflowHandle<T, E> executeWorkflow(
           if (activeWorkflows.putIfAbsent(workflowId, bucket) != null) {
             throw new DBOSWorkflowExecutionConflictException(workflowId);
           }
+          // Release the active-ID entry before the terminal outcome write becomes durable:
+          // once it is visible, a resume can re-dispatch this workflow to this executor, and
+          // a stale entry would reject that dispatch. The once-guard keeps the finally
+          // backstop from removing an entry a resumed execution re-acquired in the meantime.
+          var activeReleased = new AtomicBoolean(false);
+          Runnable releaseActive =
```
devhawk: Is there a way to use autoclosable / try with resources here? I'm worried that future updates to executeWorkflow will forget to wire this up correctly

maxdml: Added `ActiveWorkflowGuard`

transact/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
```diff
@@ -1743,10 +1771,8 @@ private <T, E extends Exception> WorkflowHandle<T, E> executeWorkflow(
               finalOptions.isDequeuedRequest()
                   ? new QueueBucket(finalOptions.queueName(), finalOptions.queuePartitionKey())
                   : NO_QUEUE;
-          if (activeWorkflows.putIfAbsent(workflowId, bucket) != null) {
-            throw new DBOSWorkflowExecutionConflictException(workflowId);
-          }
-          try {
+          var active = new ActiveWorkflowGuard(workflowId, bucket);
+          try (active) {
```
devhawk:
```suggestion
          try (var active = new ActiveWorkflowGuard(workflowId, bucket)) {
```

devhawk (approved):
Minor nit to still fix, otherwise looks good

PR #462, opened 2026-07-23
Re-stamp workflow executor_id when successfully checkpointing a step

maxdml:
When an executor "wins" the checkpointing race, it should restamp the executor ID column in the status table so reporting is accurate.

### DIFF
--- a/transact/src/main/java/dev/dbos/transact/database/SystemDatabase.java
+++ b/transact/src/main/java/dev/dbos/transact/database/SystemDatabase.java
@@ -107,14 +107,15 @@ public class SystemDatabase implements AutoCloseable {
       String schema,
       boolean created,
       DBOSSerializer serializer,
-      boolean useListenNotify) {
+      boolean useListenNotify,
+      String executorId) {
     validatePostgresDataSource(dataSource);
     schema = sanitizeSchema(schema);
     if (schema.contains("\"")) {
       throw new IllegalArgumentException("Schema name must not contain double quotes");
     }
 
-    this.ctx = new DbContext(dataSource, schema, serializer, this.closed::get);
+    this.ctx = new DbContext(dataSource, schema, serializer, this.closed::get, executorId);
     this.created = created;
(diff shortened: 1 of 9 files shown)


transact/src/main/java/dev/dbos/transact/database/SystemDatabase.java
```diff
@@ -136,32 +137,42 @@ public SystemDatabase(
...
+    return create(config, null);
+  }
+
+  public static SystemDatabase create(DBOSConfig config, String executorId) {
     if (config.dataSource() == null) {
       return new SystemDatabase(
-          config.databaseUrl(),
-          config.dbUser(),
-          config.dbPassword(),
+          createDataSource(config.databaseUrl(), config.dbUser(), config.dbPassword()),
```
maxdml, line 162: @devhawk  it looks like there might be a pre-existing issue where this path would ignore `config.useListenNotify()`

transact/src/main/java/dev/dbos/transact/database/dao/StepsDAO.java
```diff
@@ -105,6 +116,20 @@ static void recordStepResult(
...
+          """
+              .formatted(schema);
+      try (var stmt = conn.prepareStatement(updateSql)) {
+        stmt.setString(1, executorId);
+        stmt.setString(2, result.workflowId());
+        stmt.setString(3, executorId);
+        stmt.executeUpdate();
+      }
+    }
   }
```
devhawk: this seems like it should be a static method on workflowDAO

transact/src/main/java/dev/dbos/transact/database/SystemDatabase.java
devhawk: You mean in the else clause? yeah, looks like an issue. If you want to fix it here, go ahead. Otherwise please open an issue

devhawk (approved):

PR #445, opened 2026-06-19
AOT CLI

devhawk:
removes postgres & workflow CLI commands, leaving just migrate and reset

updates cli build to support native image generation + GH actions to produce linux/macos/windows cli images for publish builds (always created as build artifacts, attached to GH releases when created)

fixes #362

Issue #444, opened 2026-06-19
Annotation Processor support


devhawk:
Investigate using annotation processing to add @Workflow/@Step processing 

Note, we have a version of this for [SpringBoot](https://docs.dbos.dev/java/tutorials/spring-boot-integration#defining-workflows-and-steps) using their AOT stack, but this issue is to track leveraging [JDK annotation processing](https://docs.oracle.com/javase/8/docs/api/javax/annotation/processing/Processor.html)

hiSandog:
Annotation processor support should include both build-tool wiring and incremental-build behavior. A good fixture would compile a small project with annotated workflows/transactions under Maven and Gradle, then change one annotated class and assert generated metadata updates without requiring a clean build. That catches the common processor bug where the first full compile works but incremental IDE/build usage sees stale generated files.

PR #471, opened 2026-08-19
Support system database sharing (application names) and cross-application enqueue

### LINKED ISSUE #468: Support System Database Sharing
Analogous to https://github.com/dbos-inc/dbos-transact-py/pull/809. This should support full interop with other languages within a shared system database.

devhawk:
Closes #468. Port of [python#809](https://github.com/dbos-inc/dbos-transact-py/pull/809), [ts#1326](https://github.com/dbos-inc/dbos-transact-ts/pull/1326), and go [#444](https://github.com/dbos-inc/dbos-transact-go/pull/444)/[#445](https://github.com/dbos-inc/dbos-transact-go/pull/445).

Every DBOS object — workflows, steps, queues, schedules, application versions — records the application that owns it, so several applications can share one system database in isolation, or interoperate deliberately by naming each other's objects.

**`NULL` means unclaimed.** The columns are added nullable with no default, so every pre-existing row, and every row written by an SDK that does not know the column, belongs to all of them. Reads match `application_name = ? OR application_name IS NULL` throughout. That is what makes the migration safe, and also what makes the isolation one-directional: naming your applications hides them from each other, but not from an executor that ignores the column.

## Schema

Migrations 100–107, the start of a cross-SDK shared history: from index 100 on, every SDK defines the same migration at the same index. This language's own history (1–47) is padded out to 99, and both the runner and `dbos migrate --print` skip the padding without a round trip each, recording it in a single write.

| # | |
|---|---|
| 100–104 | `application_name` on `workflow_status`, `queues`, `workflow_schedules`, `application_versions`, `operation_outputs` |
| 105 | `enqueue_workflow` gains a trailing `application_name`; every parameter is defaulted, so callers that omit it still resolve to it |
| 106–107 | the partial unique index pair that replaces `version_name`'s global uniqueness (which is *not* dropped — not until every SDK reaching a shared database is past 107) |

## What is scoped

Listings (workflows, queues, schedules, versions, aggregates, metrics), the queue dequeue and its rate-limit and concurrency counts, the latest-version lookup, and the delayed-workflow and garbage-collection sweeps. Recovery follows from the workflow listing — which matters, because `executor_id` defaults to the literal `local` in all four SDKs, so two unnamed applications sharing a database would otherwise each treat the other's PENDING workflows as their own to recover.

Deliberately **not** scoped:

- **Anything addressed by workflow ID.** A fork is the forking application's own work, steps included.
- **The debounce holder lookup.** The deduplication index is global across the applications sharing the database, so a caller cannot steer around a holder it cannot see. The lookup returns the holder's owner alongside its ID, and a debounce against a peer's key raises `DBOSQueueDuplicatedException` rather than trying to extend it — mirroring the foreign-holder arm of Python's `_classify_bounce`. Scoping the read instead would hide the holder while the global unique index still rejected the insert, and the retry loop would spin.

Registering a queue, schedule, or application version under a name another application holds raises `DBOSApplicationNameConflictException` — names are shared address space, so a collision is not ours to resolve. Claims are written as `COALESCE(application_name, ?)`, so a registration landing between the ownership check and the write keeps the name it just took.

## Enqueueing across applications

`DBOS.enqueueWorkflow` takes the same `EnqueueOptions` as `DBOSClient` and writes the same row, so a Java application can enqueue a workflow implemented in another process or another language without holding a reference to its function. `enqueuePortableWorkflow` carries named arguments, for targets like a Python workflow with keyword arguments.

Called from inside a workflow it behaves like `startWorkflow`: the enqueued workflow is recorded as a child under a derived ID, so a crash-replay collides with the original enqueue instead of enqueueing a second workflow. Called from a step it throws, as starting a workflow from a step already does.

Two deliberate differences from `startWorkflow`: the workflow name and queue are not checked against the local registries, since neither may exist here, and the application version is left unset unless given, so the owning application's own latest version picks the work up.

`DBOS.listQueues` and `DBOS.listSchedules` gained the same application filter the client has.

## Behaviour changes worth knowing

- **The application name is hashed into the computed application version**, so two applications built from one jar do not collide on a single version row. Existing deployments therefore compute a new version on upgrade, as they do in Python, TypeScript, and Go.
- **The executor's resolved name is the one that owns rows.** On DBOS Cloud the executor takes its name from `DBOS_APP_NAME` rather than from the config; ownership, the version hash, and the peer-ownership checks all key on that single identity, as Python threads one `GlobalParams.app_name` into both.
- `DBOSClient` takes an optional application name; without one it owns nothing and sees every application's rows.
- `StepInfo` exposes the application owning each step, and export/import round-trips it, so imported steps keep exactly the owner they were exported with rather than being re-stamped.
- The conductor protocol carries `application_name` on workflow, queue, schedule, and version rows, and accepts it as a filter on the list and aggregate requests. A Conductor that predates the field sends nothing, which leaves every listing scoped to the application.

## Not included

`rename-application` (re-owning rows after a rename, or adopting unclaimed ones) is a follow-up, matching Go. Until it lands, the collision error points at the Python or TypeScript CLI, which operates on the shared system database whatever wrote the rows.

Queue and schedule deletion, and pause/resume, still address rows by name without an ownership predicate — matching Python and TypeScript, which do the same. Tightening that is cross-SDK work, not Java's to do alone.

## Testing

`ApplicationNameTest` runs two applications against one system database: stamping across all five tables, listing isolation and the explicit cross-application filter, unclaimed rows visible to both, dequeue isolation, fork ownership, the three collision errors, version scoping, a nameless client, step ownership through export and import, both debouncers refusing a peer's deduplication key, an enqueue-by-name inheriting its parent's timeout, and a contested schedule batch rolling back whole. `DebouncerHolderTest` covers holder ownership and the replay of a `DBOS.lookupDebouncer` step recorded before this column existed. `MigrationManagerTest` covers upgrading a database recorded at 47 through the padding to 107.


transact/src/main/java/dev/dbos/transact/config/DBOSConfig.java
```diff
@@ -100,6 +101,10 @@ public record DBOSConfig(
     if (appName == null || appName.isEmpty()) {
       throw new IllegalArgumentException("DBOSConfig.appName must not be null or empty");
     }
+    if (!Validation.isValidApplicationName(appName)) {
+      throw new IllegalArgumentException(
+          Validation.invalidApplicationName("DBOSConfig.appName", appName));
+    }
```
maxdml: This is actually a breaking change -- we don't do that check in Go/TS, but it looks like we do it in [Python](https://github.com/dbos-inc/dbos-transact-py/blob/9c219ad5fdd30557e00dfc8f15158f3c336c1ffc/dbos/_dbos_config.py#L579-L584). Probably OK but not cool if someone hits that landmine. Do we really need that validation?

transact/src/main/java/dev/dbos/transact/database/dao/WorkflowDAO.java
```diff
@@ -1666,7 +1744,8 @@ private static void forkWorkflows(
             applicationVersion,
             queueName,
             queuePartitionKey,
-            timeoutMs);
+            timeoutMs,
+            ctx.appName());
```
maxdml: Does that mean that the forked workflow unconditionally receives the current application's name? It should be the name of the source application -- and fall back to the current's if none exist for the row.

kraftp (approved):
Let's also add this to https://github.com/dbos-inc/dbos-demo-apps/tree/main/interops

devhawk:
> Let's also add this to https://github.com/dbos-inc/dbos-demo-apps/tree/main/interops

this PR has to merge first

PR #473, opened 2026-08-24
Reset pooled test databases with DELETE instead of TRUNCATE

devhawk:
CockroachDB implements TRUNCATE as a schema change, so it prices like CREATE INDEX however few rows a table holds. Measured against this schema it costs 1.16s where the equivalent deletes cost 0.05s — 23x — and `acquire` resets a pooled container on every `new PgContainer()`, which 46 test classes do in @BeforeEach. That is roughly 900 resets a run.

CI's own test-results artifacts say this is where the CockroachDB leg goes. Test bodies cost the same on both backends — 1397s on CockroachDB against 1435s on PostgreSQL — while per-class fixture time is +1308s on CockroachDB and ~0 on PostgreSQL. Roughly 1000s of that is these truncates.

TRUNCATE would win only once a table is big enough for row count to dominate, which no test fixture is.

The table list now comes from the catalogue rather than a hard-coded list, so a migration that adds a table cannot silently leave it uncleaned. Deleting from all of them in one statement batch is safe in any order: emptying everything cannot strand a foreign key.

### DIFF
--- a/transact/src/test/java/dev/dbos/transact/utils/PgContainer.java
+++ b/transact/src/test/java/dev/dbos/transact/utils/PgContainer.java
@@ -8,6 +8,7 @@ import dev.dbos.transact.migrations.MigrationManager;
 import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.SQLException;
+import java.util.ArrayList;
 import java.util.Objects;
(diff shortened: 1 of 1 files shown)


devhawk:
Abandoned.

PR #476, opened 2026-08-28
Compatibility fixes for system database sharing

devhawk:
Compatibility fallout from #471, which added application names and system
database sharing. Two independent kinds, in one PR because they share that
cause.

The first three commits were opened before as #475, which was closed unmerged;
they are unchanged here.

## Reading rows another application wrote

Sharing a system database means a read can land on a row this runtime did not
write, and two such reads failed outright.

**Read an empty error column as no error** (`e07f89b`). The Go SDK stores a
workflow's error as a non-nullable string, so a workflow that *succeeded*
leaves `""` where this SDK writes NULL. Java parsed that as JSON and threw
`MismatchedInputException: No content to map due to end-of-input` — on a
workflow that never failed. TypeScript and Python already tolerate both forms.
Applied where the column is read, not inside the deserializers, which should
not have to know how a column came to be written. What this SDK writes is
unchanged.

**Don't try to deserialize a format we don't recognize** (`11897bc`).
`getWorkflowStatus` threw `IllegalArgumentException: Serialization is not
available` for any row in a format this runtime has no deserializer for. A
workflow ID addresses the whole database, so a status read reaches other
applications' rows by design, and what is wanted from one is the metadata.
This is not only cross-SDK: two Java apps on one database, one with a custom
serializer, hit it identically. `SerializationUtil.canDeserialize` now asks
directly, and the reads that assemble a record from a row report unreadable
payloads as null.

Only where the payload is *one field of a record*. Where the payload is the
answer it still throws — `getEvent`, a workflow's result, a recv'd message, a
stream, and a recorded step result on replay. That last one matters most: a
step that "returned null" because its output could not be read would corrupt
the run it is replaying, so it has a test of its own. Three paths read a record
but must not accept a null payload and check for themselves: running a workflow
(refusing rather than invoking it with arguments it never had), exporting one,
and importing one.

## Restoring constructors #471 broke

#471 added `applicationName` to a number of public records. For a Java record
that changes the *canonical constructor in place*, so every caller that built
one positionally stopped compiling — including the DBOS conductor test app,
whose CI cannot build against the published SDK
([failing run](https://github.com/dbos-inc/dbos-conductor/actions/runs/33212895492)):

```
App.java:277: error: no suitable constructor found for WorkflowSchedule(...)
```

**Restore constructors that omit the application name** (`2a7aeeb`) —
`WorkflowSchedule`, `Queue`.
**Restore input constructors that omit the application name** (`68297a2`) —
`ListWorkflowsInput`, `GetStepAggregatesInput`, `GetWorkflowAggregatesInput`.
**Restore the EnqueueOptions constructor that omits the application name**
(`3054cc6`) — `DBOSClient.EnqueueOptions`.

Each takes the pre-#471 argument list and forwards a null owner, which is the
same default the existing no-arg constructors already use: the creating or
enqueueing application for the records that record an owner, and "this
application's rows plus unclaimed ones" for the filters. `GetWorkflowAggregatesInput`
needed care — its two new components landed *mid-list* rather than appended, so
that delegation reorders its arguments and passes `groupByApplicationName` as
false.

### Scope

A sweep of every public signature #471 changed (1,614 → 1,678 across all six
published modules) found 30 changed in place. They fall out as:

| | Disposition |
|---|---|
| The 6 records above | Constructors restored |
| `WorkflowStatus`, `StepInfo`, `VersionInfo` | Left alone — callers receive these, they do not construct them |
| 21 internal signatures | Left to change in place — conductor wire DTOs, the DAO layer, `DbContext`, `SystemDatabase`, `DBOSExecutor`, and the `internal` packages |

Internal APIs deliberately get no compatibility overloads; they are fixed at
their call sites when they change. Only external app code, which cannot be
fixed that way, gets them.

No public interface, enum, or abstract class was modified, so there are no
implementor breaks, and no record component was removed or renamed — every
change was an addition. `transact-cli` only gained a new class.

### Testing

The overloads delegate positionally, where a mis-ordered argument would compile
cleanly and silently scramble fields, so each is compared against the canonical
constructor rather than spot-checked. Both new test classes are container-free.
Verified non-vacuous by mutation: swapping two same-typed arguments in the
`GetWorkflowAggregatesInput` and `EnqueueOptions` delegations makes the
corresponding test fail.

Note that this does not unblock conductor CI on its own — the test app resolves
`dev.dbos:transact:+`, so it needs a new prerelease published.

### DIFF
--- a/transact/src/main/java/dev/dbos/transact/DBOSClient.java
+++ b/transact/src/main/java/dev/dbos/transact/DBOSClient.java
@@ -371,6 +371,49 @@ public class DBOSClient implements AutoCloseable {
           null);
     }
 
+    /**
+     * Constructs options with no explicit owning application, which enqueues the workflow for the
+     * enqueueing application.
+     */
+    public EnqueueOptions(
+        @NonNull String workflowName,
+        @Nullable String className,
+        @Nullable String instanceName,
+        @NonNull String queueName,
+        @Nullable String workflowId,
+        @Nullable String appVersion,
+        @Nullable Duration timeout,
+        @Nullable Instant deadline,
+        @Nullable String deduplicationId,
+        @Nullable Integer priority,
+        @Nullable String queuePartitionKey,
+        @Nullable Duration delay,
+        @Nullable SerializationStrategy serialization,
+        @Nullable String authenticatedUser,
+        @Nullable String assumedRole,
+        @Nullable List<String> authenticatedRoles,
+        @Nullable Map<String, Object> attributes) {
+      this(
+          workflowName,
+          className,
+          instanceName,
+          queueName,
+          workflowId,
+          appVersion,
+          timeout,
+          deadline,
+          deduplicationId,
+          priority,
+          queuePartitionKey,
+          delay,
+          serialization,
+          authenticatedUser,
+          assumedRole,
+          authenticatedRoles,
+          attributes,
+          null);
+    }
+
     public EnqueueOptions(
         @NonNull String workflowName, @Nullable String className, @NonNull String queueName) {
       this(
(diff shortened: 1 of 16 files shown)


kraftp (approved):
Changes look good!

PR #493, opened 2026-09-12
Stop recovery adopting workflows this executor is already running

### LINKED ISSUE #477: CancelResumeRaceTest.activeIdReleasedBeforeOutcomeWriteTest is flaky (expected 2 but was 3)
`CancelResumeRaceTest.activeIdReleasedBeforeOutcomeWriteTest` failed on one
matrix entry of an otherwise green run, on a branch that does not touch it.

**Run:** [33219043815](https://github.com/dbos-inc/dbos-transact-java/actions/runs/33219043815) (PR #476), job `test / test (25, temurin)`

```
org.opentest4j.AssertionFailedError: expected: <2> but was: <3>
	at dev.dbos.transact.workflow.CancelResumeRaceTest.activeIdReleasedBeforeOutcomeWriteTest(CancelResumeRaceTest.java:170)
```

### Why this looks like flakiness rather than a defect

- The same commit passed the same suite on the five other matrix entries in
  that run: 17/temurin, 21/temurin, 17/oracle, 21/oracle, **25/oracle**. Only
  25/temurin failed, so it is not a JDK 25 language or API problem.
- The PR is additive constructors plus two read-path changes, and does not
  touch this test, the queue service, or dispatch. The test has not been
  modified since it was introduced in #456.
- The failure is a count being one too high, not a timeout or an assertion
  about correctness of the thing under test.

### The assertion being raced (hypothesis)

Line 170 asserts the workflow body ran **exactly** twice:

```java
assertTrue(impl.secondRunDone.await(15, TimeUnit.SECONDS), ...);
assertEquals("completed", resumedHandle.getResult());
assertEquals(2, impl.runs.get());   // <-- line 170
```

`secondRunDone` is counted down by *any* run after the first, so it tells the
test that a second dispatch happened — not that no third one will. The test
runs with `setSpeedupForTest()` on the queue service, and reaches this line
while run 1's terminal outcome write is still parked by the JDBC proxy, which
is exactly the window in which the row is dispatchable. If the poller dispatches
once more before the assertion reads the counter, `runs` is 3.

So the count is bounded by poller timing rather than by a barrier, which would
explain an occasional extra dispatch under different scheduling. I have not
confirmed this by reproducing it — it is the reading that fits the evidence, and
whether a third dispatch in that window is itself acceptable behaviour is the
part worth a maintainer's eye. If it is, the fix is on the test side (assert no
*further* dispatch, rather than an exact count observed at an arbitrary moment).
If it is not, the flake is pointing at something real in dispatch.

[CancelResumeRaceTest.java#L148-L175](https://github.com/dbos-inc/dbos-transact-java/blob/aa9a7ca4779b62b3f7fc4b484ead401daeb5fc02/transact/src/test/java/dev/dbos/transact/workflow/CancelResumeRaceTest.java#L148-L175)

First observation as far as I can tell; there is no prior issue for this test.
Filing it so a recurrence has somewhere to land. Compare #474, which tracks a
different flaky test.

### LINKED ISSUE #488: WorkflowOutcomeOwnershipTest.failedRunFailsFastWhenItsRowIsDeleted is flaky (reported its own error instead of DBOSNonExistentWorkflowException)
`WorkflowOutcomeOwnershipTest.failedRunFailsFastWhenItsRowIsDeleted` failed once in CI on an unrelated PR ([#487](https://github.com/dbos-inc/dbos-transact-java/pull/487), a Gradle wrapper bump), in the `test / test (21, oracle)` job only — the other five JDK/vendor jobs and `test-crdb` passed.

[Failing job](https://github.com/dbos-inc/dbos-transact-java/actions/runs/34644734290/job/103412671461):

```
WorkflowOutcomeOwnershipTest > failedRunFailsFastWhenItsRowIsDeleted() FAILED
    org.opentest4j.AssertionFailedError: a failed run whose row vanished must not report its own error or poll forever ==> Unexpected exception type thrown, expected: <dev.dbos.transact.exceptions.DBOSNonExistentWorkflowException> but was: <java.lang.IllegalStateException>
        at dev.dbos.transact.workflow.WorkflowOutcomeOwnershipTest.failedRunFailsFastWhenItsRowIsDeleted(WorkflowOutcomeOwnershipTest.java:341)

        Caused by:
        java.lang.IllegalStateException: own failure
            at dev.dbos.transact.workflow.OutcomeOwnershipServiceImpl.throwingWorkflow(WorkflowOutcomeOwnershipTest.java:440)
```

### Why this may be more than test flakiness

The test is strictly ordered, with no timing window of its own:

1. `startThrowingRun` blocks until the workflow body has counted down its started latch, so the row is inserted and the run is parked.
2. `deleteRow` deletes the row and asserts `executeUpdate() == 1`, so the row definitely existed and was definitely removed.
3. Only then does `releaseRun` let the workflow throw `IllegalStateException("own failure")`.

The run's error write is `UPDATE ... SET status = ... WHERE workflow_uuid = ? AND status = 'PENDING'` (`WorkflowDAO.updateWorkflowOutcome`), so against a deleted row it must match 0 rows, `persistWorkflowError` must return `false`, and `DBOSExecutor.executeWorkflow` must take the adopt path — `awaitWorkflowResult(workflowId, true)` — which surfaces `DBOSNonExistentWorkflowException`.

Surfacing the run's own `IllegalStateException` instead means `persistWorkflowError` returned `true`: the UPDATE matched a PENDING row that had already been deleted and committed by the test's connection. That is a visibility/ordering anomaly rather than an obviously benign race in the test, so it may point at a real bug in the outcome-ownership path (or at `dbRetry` masking a failed attempt as a successful write).

### History

- The test was added 2026-08-21 in #467 (`a623ae1`).
- This is its first observed failure: every other failed CI run since then was `javadoc`, `spotlessJavaCheck`, or an infrastructure/timeout failure with no failing test assertion.
- Not reproduced locally yet; no known frequency.

### LINKED ISSUE #490: DBOSExecutorTest.sleep is flaky: asserts sleep round-trip completes within 400ms of the sleep duration
`DBOSExecutorTest.sleep` asserts an upper bound on how long a 2-second `DBOS.sleep` takes to come back, with 400ms of headroom:

```java
long start = System.currentTimeMillis();
try (var id = new WorkflowOptions(wfid).setContext()) {
  executingService.sleepingWorkflow(2);
}

long duration = System.currentTimeMillis() - start;
logger.info("Duration {}", duration);
assertTrue(duration >= 2000);
assertTrue(duration < 2400); // Relaxed a bit for CI
```

[`DBOSExecutorTest.java:291`](https://github.com/dbos-inc/dbos-transact-java/blob/main/transact/src/test/java/dev/dbos/transact/execution/DBOSExecutorTest.java#L291)

### Occurrence

2026-08-27, PR #471 — [`test (21, temurin)`](https://github.com/dbos-inc/dbos-transact-java/actions/runs/33101622832/job/98620627843):

```
DBOSExecutorTest > sleep() FAILED
    org.opentest4j.AssertionFailedError: expected: <true> but was: <false>
        at dev.dbos.transact.execution.DBOSExecutorTest.sleep(DBOSExecutorTest.java:291)
```

The surrounding log shows the job mid-`ScaleServiceImpl` run, thousands of workflows deep, so the 400ms of slack was being spent on database round trips and scheduler wake-up latency rather than on anything sleep-related.

### Why it is the wrong assertion

The upper bound measures the per-workflow round trip — launch, checkpoint the `DBOS.sleep` step, poll, resume, write the outcome — not the sleep. A slow database pushes it over the line while the behaviour under test is perfectly correct. This is the same shape as the two assertions fixed in #482 (`Fixes #481 and #474`), which replaced latency upper bounds with behavioural lower bounds; `DBOSExecutorTest.sleep` was not covered by that pass.

The lower bound (`duration >= 2000`) is the one that actually tests something: the sleep must not return early. The steps assertion below it (`DBOS.sleep` was checkpointed) covers the rest.

### Suggested fix

Drop the `duration < 2400` assertion, or widen it to something that could only fail if the sleep were being re-run (e.g. `< 2 * 2000`).

### LINKED ISSUE #491: Launch-time recovery can re-execute a workflow that is already running in-process (duplicate execution)
`AsyncWorkflowTest.sameWorkflowId` failed on a push to `main` with the workflow body having executed twice:

```
AsyncWorkflowTest > sameWorkflowId() FAILED
    org.opentest4j.AssertionFailedError: expected: <1> but was: <2>
        at dev.dbos.transact.workflow.AsyncWorkflowTest.sameWorkflowId(AsyncWorkflowTest.java:58)
```

[`test (25, oracle)`](https://github.com/dbos-inc/dbos-transact-java/actions/runs/32182419377/job/95858322794), 2026-08-18, on `main` (`Re-stamp workflow executor_id when successfully checkpointing a step`). `executionCount` is bumped once per `workWithString` invocation (`SimpleService.java:79`), and the test's calls are strictly sequential, so the count can only reach 2 if the body ran twice for a single workflow ID.

### What the log shows

Launch-time recovery adopted `wf-123` while the caller was still running it, and ran the body a second time:

```
20:32:37.823 [ForkJoinPool-1-worker-9] DBOS Executor started
20:32:37.823 [ForkJoinPool-1-worker-9] runWorkflow workWithString/...
20:32:37.823 [ForkJoinPool-1-worker-9] initWorkflowStatus workflowId wf-123
20:32:37.823 [ForkJoinPool-1-worker-9] insertWorkflowStatus workflowId wf-123
20:32:37.826 [virtual-21805]           executeWorkflow task workWithString/... workflowId=wf-123
20:32:37.826 [virtual-21805]           Executed workflow workWithString          <-- run #1
20:32:37.827 [virtual-21804]           recoverWorkflow execute wf-123
20:32:37.827 [virtual-21804]           executeWorkflowById wf-123
20:32:37.828 [virtual-21804]           executeWorkflow workWithString/... workflowId=wf-123
20:32:37.828 [virtual-21804]           initWorkflowStatus workflowId wf-123
20:32:37.829 [virtual-21804]           insertWorkflowStatus workflowId wf-123
20:32:37.830 [virtual-21805]           updateWorkflowOutcome wfid wf-123 status SUCCESS
20:32:37.831 [virtual-21806]           executeWorkflow task workWithString/... workflowId=wf-123
20:32:37.831 [virtual-21806]           Executed workflow workWithString          <-- run #2
20:32:37.834 [virtual-21806]           updateWorkflowOutcome wfid wf-123 status SUCCESS
```

Four milliseconds after `DBOS Executor started`, the test calls `workWithString` with `workflowId=wf-123`, which inserts a `PENDING` row. The recovery pass that `launch()` kicks off then scans for `PENDING` workflows belonging to this executor, finds that row, and calls `executeWorkflowById` on it — while the original invocation is still inside the workflow body, before it has written its outcome. Both copies run to completion and both write `SUCCESS`.

### Why this matters beyond the test

This is the once-and-only-once guarantee being violated, not a timing assertion:

- Recovery has no guard against adopting a workflow that is **currently executing in this same process**. A `PENDING` row that belongs to a live in-process run is indistinguishable, to the recovery scan, from one orphaned by a crash.
- The window is only as narrow as it is here because the test starts a workflow immediately after `launch()`. In an application that enqueues or starts work as soon as it launches — which is the common shape — the same window is open for whatever the startup recovery sweep takes to complete.
- Nothing downstream caught it. Both runs wrote `SUCCESS` over each other, so the workflow looks correct in the database; only the side effect (the counter, or in a real app whatever the workflow does) betrays the double execution.

### Suggested direction

Recovery should skip workflow IDs that the executor knows it is currently running — the executor already tracks in-flight workflows for cancellation, so the set is available — or the recovery sweep should be ordered against the point at which `launch()` begins accepting new work, so that it can only see rows that predate this process.

Note `SyncWorkflowTest.sameWorkflowId` (`SyncWorkflowTest.java:109`) asserts the same invariant, and has the same exposure.

devhawk:
Fixes #488, fixes #491. Very likely fixes #477 (see caveat below). Fixes #490, which is unrelated to the rest.

## One defect behind three reports

#477, #488 and #491 are the same bug: the launch-time recovery sweep adopting a workflow that this process started microseconds earlier, and running its body a second time.

`start()` bounds the sweep with `withEndTime(Instant.now())`, read at the *end* of launch, and the filter is `created_at <= endTime` at millisecond granularity. There is nothing between that read and the caller's first `startWorkflow`, so the two land in the same millisecond and the just-inserted row is `<=` the cutoff. From the #488 job:

```
20:35:25.577  DBOS Executor started
20:35:25.577  startWorkflow ... insertWorkflowStatus outcome-ownership-failed-deleted-...
20:35:25.579  executeWorkflow task throwingWorkflow ...        <- run A (the test's)
20:35:25.580  recoverWorkflow execute outcome-ownership-...
20:35:25.581  insertWorkflowStatus outcome-ownership-...       <- row re-inserted
20:35:25.583  executeWorkflow task throwingWorkflow ...        <- run B (recovery's)
20:35:25.583  updateWorkflowOutcome ... status ERROR           <- run A
20:35:25.584  updateWorkflowOutcome ... status ERROR           <- run B
```

Both copies run to completion and race to record the outcome. What each test noticed differed:

- **#491** (`AsyncWorkflowTest.sameWorkflowId`) — the body simply ran twice, so `executionCount` was 2.
- **#488** (`WorkflowOutcomeOwnershipTest.failedRunFailsFastWhenItsRowIsDeleted`) — recovery's re-insert resurrected the row the test had deleted, so the live run's `UPDATE ... WHERE status = 'PENDING'` matched after all, it owned the outcome, and it reported its own `IllegalStateException` instead of failing fast. That is why #488 read as a visibility anomaly; it wasn't one.
- **#477** (`CancelResumeRaceTest`) — a recovery duplicate of run 1 makes the resumed dispatch the *third* run, which is the `expected: <2> but was: <3>` reported.

`ActiveWorkflowGuard` does not stop this. It is released just before the terminal outcome write, so a recovery landing in that window acquires the ID and re-runs the body.

## The change

**The sweep's cutoff moves to the top of `start()`, before the `isRunning` gate.** Everything `start()` goes on to do — the app-version hash over every registered workflow, the executor service, migrations, queue and scheduler start, listener callbacks — now lands between the read and the earliest a caller can start a workflow.

The one-millisecond step-back stays, doing a different job: placement buys margin, the step-back makes the bound *exclusive*. `created_at` is written from `System.currentTimeMillis()` (the same clock this reads), so `<= cutoff - 1ms` is exactly `< cutoff` on a millisecond-granular column. Without it a `start()` that finishes inside a millisecond — already-migrated database, no queues, no listeners — reopens the same hole.

**`recoverWorkflow` skips a workflow whose ID is active on this executor.** `recoverPendingWorkflows()`, the Conductor/AdminServer entry point, carries no time bound at all, so this is the only thing standing between an operator-triggered recovery and a duplicate execution.

### What this does not fix

The launch path is now closed deterministically: the sweep can only see workflows created strictly before this process started, and none of those is running in it.

The operator-triggered path is narrowed, not closed. The active-ID check is a sample, not a lock, and during the `ActiveWorkflowGuard` release window the ID is absent from the map — so a recovery landing exactly there still passes the check. Closing that means changing the release-before-write ordering, which is the invariant `CancelResumeRaceTest` exists to pin down, and is a design call rather than a flaky-test fix. Left alone deliberately.

## Separately: #490

`DBOSExecutorTest.sleep` asserted a 2s sleep returns within 2400ms. That measures the round trip around the sleep — launch, checkpoint, poll, resume, outcome write — not the sleep, so it reddens on whichever job is slowest. Widened to a bound only a repeated sleep could break, the same correction #482 made to `ChaosTest` and the queue tests, which missed this one.

## Testing

`RecoveryServiceTest.recoveryLeavesAWorkflowThisExecutorIsRunningAlone` drives recovery directly, since the millisecond collision that produced these failures is not something a test can schedule. Confirmed it fails without the `recoverWorkflow` guard — recovery re-inserts the deleted row — and passes with it.

33 tests pass across `RecoveryServiceTest`, `WorkflowOutcomeOwnershipTest`, `AsyncWorkflowTest`, `CancelResumeRaceTest` and `DBOSExecutorTest`; `spotlessCheck`, `compileTestJava` and `javadoc` clean.

I have left #477 for CI to confirm rather than closing it on reasoning: its original run logs have expired, so the link to this root cause is inferred from the test's shape (`setUp()` launches, `runCancelAndPark()` starts a workflow immediately) rather than observed. Worth a few green runs before trusting it.


transact/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
```diff
@@ -1207,6 +1212,14 @@ public boolean deprecatePatch(String patchName) {
...
     Objects.requireNonNull(workflowId, "workflowId must not be null");
 
+    // A workflow running here is not orphaned, whatever its row says. recoverPendingWorkflows()
+    // has no time bound, so this is the only thing stopping an operator-triggered recovery from
+    // starting a second execution of a live one.
+    if (activeWorkflows.containsKey(workflowId)) {
+      logger.debug("recoverWorkflow skip active {}", workflowId);
+      return retrieveWorkflow(workflowId);
+    }
+
```
maxdml: Recovering an active workflow wouldn't cause a 2x execution, right? Because `executeWorkflowById` should already do the check above?

kraftp (approved):

transact/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
devhawk: Added an `activeWorkflows` check in `recoverWorkflow`. `executeWorkflowById` does not check active workflows directly. Prior to this change, the only use of `activeWorkflows` to manage workflow execution is in the workflow task in `executeWorkflow`.

maxdml: Got it. My misunderstanding came before realizing that Java does not re-enqueue recovery calls (otherwise it'd go through the same dequeue path that checks activeWorkflows IIUC)

PR #497, opened 2026-09-14
VMID fixes

maxdml:
- Set to local when DBOS__VMID is set but empty
- Don't regenerate an executor ID when a conductor key is provided in the code config and we are in DBOS Cloud


transact/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java
```diff
@@ -185,7 +185,8 @@ public DBOSExecutor(DBOSConfig config) {
             ? Objects.requireNonNullElse(System.getenv("DBOS_APP_NAME"), "")
             : config.appName();
     appVersion = Objects.requireNonNullElse(System.getenv("DBOS__APPVERSION"), "");
-    executorId = Objects.requireNonNullElse(System.getenv("DBOS__VMID"), "local");
+    var envExecutorId = System.getenv("DBOS__VMID");
+    executorId = envExecutorId == null || envExecutorId.isEmpty() ? null : envExecutorId;
```
devhawk: is there a reason we set to null instead of local and then add the extra check at the end?

devhawk (approved):

PR #510, opened 2026-09-15
Scale the limiter span assertion by a measured unthrottled baseline

### LINKED ISSUE #474: DynamicQueuesTest.testLimiter is flaky on CockroachDB
`DynamicQueuesTest.testLimiter()` failed on the `test-crdb` job while every Postgres matrix job passed, on [run 33113368740](https://github.com/dbos-inc/dbos-transact-java/actions/runs/33113368740/job/98661683063) (PR #471):

```
DynamicQueuesTest > testLimiter() FAILED
    org.opentest4j.AssertionFailedError: Wave 0: Tasks 0 and 1 should start close together. Diff: 1.452 ==> expected: <true> but was: <false>
        at dev.dbos.transact.queue.DynamicQueuesTest.testLimiter(DynamicQueuesTest.java:695)

950 tests completed, 1 failed, 22 skipped
```

The test is a plain `@Test` with no CRDB gating, so it also ran in all six `test / test (17|21|25, temurin|oracle)` jobs of the same run and passed in every one. Same code, same assertion — only the slower database missed the deadline. The PR's diff touches nothing in the queue, dequeue, or rate-limiting paths.

### Why the assertion is tight

The queue is registered with `rateLimit(5, 1.8s)`, `concurrency(1)`, `workerConcurrency(1)`, so the five tasks in a wave run **one at a time**. The intra-wave assertion then requires each task to start within `waveTolerance = 1.0s` of the previous one, which really measures how fast a single workflow completes and the next is dequeued — a per-workflow round trip against the system database.

That leaves a narrow window. The inter-wave assertion (`DynamicQueuesTest.java:704-719`) requires the gap between wave starts to fall in `1.8 ± 0.5s`, i.e. at least 1.3s, while the intra-wave gap must stay under 1.0s. Only 300ms separates the intra-wave ceiling from the inter-wave floor, and CRDB's commit latency is enough to close it. The observed 1.452s is a slow round trip, not a rate limiter behaving incorrectly.

### Possible fixes

- Assert on the rate limiter's own accounting rather than observed wall-clock start times, which removes the dependency on database latency entirely.
- Or scale `waveTolerance` for CRDB, keeping enough headroom below the 1.3s inter-wave floor for the assertion to still mean something.

Raising `waveTolerance` alone is not enough: past ~1.3s the two assertions overlap and the intra-wave check stops distinguishing "same wave" from "next wave".

devhawk:
`testLimiter` bounds the span of all 15 starts by a fixed `(numWaves + 1)` periods — 7.2s. With `concurrency(1)` and `workerConcurrency(1)` the tasks run strictly one at a time, so that span is dominated by the per-workflow round trip rather than by the limiter, and CockroachDB's commit latency was enough to blow through it: 25.763s against the 7.200s bound, or ~1.7s per task. That is a slow database, not a limiter releasing at half its configured rate.

The lower-bound half of the test is fine as it stands — a slow database pushes the sliding-window gap *away* from the boundary. Only the upper bound conflates limiter delay with database cost.

## Change

Before asserting, measure the unthrottled per-task cost: run `limit` (5) of the same workflows on a second queue with identical `concurrency(1)` / `workerConcurrency(1)` and no rate limit, and take the gap between the first and last start divided by `limit - 1`. Dividing gaps rather than using the total deliberately excludes the queue's initial dispatch latency, which the rate-limited run has already paid by the time its first task starts.

The bound becomes `numWaves * periodSec + 2 * taskCost * (numTasks - 1)`: one period of jitter allowance, plus a factor of two on the measured cost. It degrades with database speed instead of assuming Postgres.

Both tests already call `setSpeedupForTest()`, which shrinks the queue polling interval 100x (it does not touch the limiter period). That field is shared across every queue listener in the process, so the baseline queue runs under the same ~10ms polling as the rate-limited one — the measurement is apples-to-apples, and `taskCost` reflects the database round trip rather than polling latency.

The failure message now also reports the measured `taskCost`, so a future failure says which of the two costs blew the bound.

## Calibration

On Postgres `taskCost` is ~0.019s, giving a 5.9s bound against an observed 3.8s span. A limiter running at half its configured rate would take 7.2s, so the assertion still catches it.

Test-only; applied identically to `DynamicQueuesTest` and `StaticQueuesTest`. `StaticQueuesTest` registers `baselineQueue` up front with the other static queues, `DynamicQueuesTest` registers it inline at the point of use, matching each test's existing style.

Fixes #474


kraftp (approved):

devhawk:
I reverted this to draft status because I'm not yet happy with the test fix. This isn't blocking a release, so there's no rush on this work.

PR #513, opened 2026-09-16
Don't let a dead-lettered workflow abort recovery

### LINKED ISSUE #461: Recovery should not be interrupted by a DLQ workflow
When recovering workflows, the entire recovery run is aborted if one of the recovered workflows gets a DLQ error.

The DLQ transition itself throws `DBOSMaxRecoveryAttemptsExceededException` after committing `MAX_RECOVERY_ATTEMPTS_EXCEEDED`:
https://github.com/dbos-inc/dbos-transact-java/blob/221ceb11a02370cbf3336d1c22397bc968775690/transact/src/main/java/dev/dbos/transact/database/dao/WorkflowDAO.java#L144-L162

Neither recovery path catches it per-workflow:

- Startup recovery wraps the entire loop in a single catch, so one throw abandons all remaining pending workflows:
https://github.com/dbos-inc/dbos-transact-java/blob/221ceb11a02370cbf3336d1c22397bc968775690/transact/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java#L267-L279

- The public `recoverPendingWorkflows` (used by Conductor) maps over the stream with no per-element catch, so the exception aborts the whole run and surfaces to the caller:
https://github.com/dbos-inc/dbos-transact-java/blob/221ceb11a02370cbf3336d1c22397bc968775690/transact/src/main/java/dev/dbos/transact/execution/DBOSExecutor.java#L1117-L1128

For reference, Python (`_recovery.py` `recover_pending_workflows`) and TypeScript (`dbos-executor.ts` `recoverPendingWorkflows`) catch per-workflow, log, and continue — a dead-lettered workflow never aborts recovery of the rest.

devhawk:
A workflow that exhausts its recovery attempts is dead-lettered by the status transition itself, which throws `DBOSMaxRecoveryAttemptsExceededException` after committing `MAX_RECOVERY_ATTEMPTS_EXCEEDED`. Neither recovery path caught it per workflow:

- startup recovery wrapped the whole loop in a single catch, so one throw abandoned every remaining pending workflow;
- `recoverPendingWorkflows` (the Conductor-facing call) mapped over the stream with no catch at all, so the exception aborted the run and surfaced to the caller.

Both paths now catch per workflow, log, and continue, matching Python's `recover_pending_workflows` and TypeScript's `recoverPendingWorkflows`. `recoverPendingWorkflows` returns handles only for the workflows it actually recovered, as Python does.

## Test

`RecoveryServiceTest.recoveryContinuesPastADeadLetteredWorkflow`: three pending workflows, the middle one's `recovery_attempts` pushed past the default limit so its recovery transition dead-letters and throws. It asserts the other two are recovered and succeed, and that the doomed one is left in `MAX_RECOVERY_ATTEMPTS_EXCEEDED`.

Verified both directions: with the fix all 6 `RecoveryServiceTest` tests pass; with the main-code change reverted the new test fails and the other 5 pass.

Fixes #461


maxdml:
We should re-enqueue on recovery rather than fixing the old path

kraftp (approved):

devhawk:
I think this PR is going to be subsumed by the PR for #514

devhawk:
The fix for #514 will make this change moot

PR #524, opened 2026-09-18
Use the configured serializer for Conductor-triggered and backfilled schedules

### LINKED ISSUE #523: Conductor trigger_schedule / backfill_schedule ignore the configured DBOSSerializer, making the workflow output unreadable
## Summary

`trigger_schedule` and `backfill_schedule` requests coming from Conductor persist the scheduled workflow with the built-in Jackson serializer (`workflow_status.serialization = java_jackson`) even when the application configured a custom `DBOSSerializer`. Since a workflow's output is written and later read in the format recorded on its row, the results of those workflows are unreadable for any value Jackson can write but not read back — e.g. Clojure's immutable maps.

The in-process `DBOS.triggerSchedule` / `DBOS.backfillSchedule` and the cron `SchedulerService` are not affected; only the two Conductor handlers are.

## Where

`dev.dbos.transact.conductor.Conductor` (1.0.0, still the same on `main`):

```java
// handleBackfillSchedule
DBOSExecutor.backfillSchedule(request.schedule_name, start, end, conductor.systemDatabase, null);
// handleTriggerSchedule
DBOSExecutor.triggerSchedule(request.schedule_name, conductor.systemDatabase, null);
```

`null` is the `DBOSSerializer serializer` parameter. `SerializationUtil.serializeArgs` falls back to `DBOSJavaSerializer` when the custom serializer is null, so `persistWorkflow` records `java_jackson`; `executeWorkflow` then serializes the result with the row's format, and `deserializeValue(..., "java_jackson")` is used to read it back.

## Reproduction (dbos-transact-java 1.0.0, custom Transit serializer via dbos-clj)

1. Register a workflow with a schedule and connect the executor to Conductor.
2. Backfill or trigger the schedule from Conductor.
3. `workflow_status` for the new rows:

```
serialization | inputs                                                       | output
java_jackson  | ["[Ljava.lang.Object;",[["java.time.Instant","2026-09-17T05:00:00Z"],null]] | {"@class":"clojure.lang.PersistentArrayMap",":dispatched":["java.lang.Long",1],":failed":["java.lang.Long",0]}
```

Every other workflow of the same application has `serialization = transit_json_verbose`.

4. `get_workflow` with `load_output: true` (or `DBOS.retrieveWorkflow(id).getResult()`) fails:

```
tools.jackson.databind.DatabindException: (was java.lang.UnsupportedOperationException)
 at [Source: REDACTED ...] (through reference chain: clojure.lang.PersistentArrayMap[":dispatched"])
  clojure.lang.APersistentMap/put at APersistentMap.java:383
  tools.jackson.databind.deser.jdk.MapDeserializer/_readAndBindStringKeyMap at MapDeserializer.java:671
  ...
```

Jackson can instantiate `PersistentArrayMap` but cannot `put` into it.

## Fix

Pass `conductor.systemDatabase.serializer()` in both handlers, the same serializer `DBOSExecutor` passes for the in-process APIs. PR incoming with regression tests (`ConductorTest.triggerScheduleUsesConfiguredSerializer` / `backfillScheduleUsesConfiguredSerializer`, which fail on `main` with `expected: <kryo> but was: <java_jackson>`).

Related: #483 (step errors written with a custom serializer could not be read back).

ovistoica:
Fixes #523.

## Summary

The Conductor handlers for `trigger_schedule` and `backfill_schedule` called the static `DBOSExecutor.triggerSchedule` / `backfillSchedule` with `null` as the custom serializer, so the workflow row was persisted with the built-in Jackson serializer (`serialization = java_jackson`) regardless of the application's configured `DBOSSerializer`. The in-process `DBOS.triggerSchedule` / `DBOS.backfillSchedule` APIs and the cron `SchedulerService` already use the configured one.

Because a workflow's output is written and read back in the format recorded on its row, the result of those workflows was unreadable for values Jackson can serialize but not deserialize (Clojure's immutable maps, for instance): `get_workflow` with `load_output`, `getResult` and `retrieveWorkflow` all failed with a `DatabindException`.

Both handlers now pass `conductor.systemDatabase.serializer()`.

## Test plan

- New `ConductorTest.triggerScheduleUsesConfiguredSerializer` and `backfillScheduleUsesConfiguredSerializer` stub `SystemDatabase.serializer()` with the test `KryoSerializer` and assert the `WorkflowStatusInternal` handed to `initWorkflowStatus` carries `serialization = kryo`. Without the fix they fail with `expected: <kryo> but was: <java_jackson>`.
- `./gradlew :transact:test --tests dev.dbos.transact.conductor.ConductorTest` — 115 tests, 0 failures.
- Observed the failure end to end against a dbos-clj (Transit serializer) executor on 1.0.0: Conductor-backfilled runs were persisted as `java_jackson` while every other run of the app was `transit_json_verbose`, and `get_workflow` failed on their output (details in #523).


devhawk:
Thanks @ovistoica for the issue and the PR. While this PR fixes #523, it revealed a deeper issue with how we handled custom  serializers. Take a look at #525 when you get a chance.

